PR #2, opened 2019-08-07
Base class - support INotifyPropertyChanged interface

wmrutten:
Added explicit interface declaration


marcovisserFurore:
This PR solves #2

PR #6, opened 2019-09-23
Feature/fhir client refactoring common

gstoian:
Moved stuff into Support.Poco. Refactored the Requester to work without knowing about Pocos. Added new entities for Request/Response when using the Requester.


mmsmits:
replaced with: https://github.com/FirelyTeam/fhir-net-common/pull/48

PR #7, opened 2019-10-01
#1071 FhirParser skips valid element

gstoian:
Added an extra check for FhirXmlNode. When PermissiveParsing is enabled for xmls parsing, the Value elements are not ignored anymore. This happened because of xmlns being not specified in the xml root and then the XElement "Value" was seen as XAttribute "Value".  

`if(scan is XElement)` is not enough (or better said it's too restrictive) as we could have the `<extension url =.. >` and then the url would be skipped.

Beside #1071, this PR should also solve the other items: #1121, #1131


marcovisserFurore:
This PR solves issue https://github.com/FirelyTeam/fhir-net-api/issues/1071

PR #8, opened 2019-10-04
1130 Expression Equals fails

gstoian:
Fix: checking the types of the 2 Expressions is needed in order to avoid a positive result when comparing different types.


marcovisserFurore:
This PR solves issue https://github.com/FirelyTeam/fhir-net-api/issues/1130

PR #11, opened 2019-10-21
1128 FhirPath expression node constructors should call its IEnumerable args only once

gstoian:
#1128
Changed the Argument type from IEnumerable to array avoiding multiple evaluation.

### DIFF
--- a/src/Hl7.FhirPath/FhirPath/Expressions/ExpressionNode.cs
+++ b/src/Hl7.FhirPath/FhirPath/Expressions/ExpressionNode.cs
@@ -118,7 +118,7 @@ namespace Hl7.FhirPath.Expressions
         public Expression Focus { get; private set; }
         public string FunctionName { get; private set; }
 
-        public Expression[] Arguments { get; private set; }
+        public IEnumerable<Expression> Arguments { get; private set; }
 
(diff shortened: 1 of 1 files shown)


src/Hl7.FhirPath/FhirPath/Expressions/ExpressionNode.cs
```diff
@@ -112,13 +112,13 @@ public FunctionCallExpression(Expression focus, string name, TypeInfo type, IEnu
...
             FunctionName = name;
-            Arguments = arguments ?? throw Error.ArgumentNull("arguments");
+            Arguments = arguments != null ? arguments.ToArray() : throw Error.ArgumentNull("arguments");
         }
 
         public Expression Focus { get; private set; }
         public string FunctionName { get; private set; }
 
-        public IEnumerable<Expression> Arguments { get; private set; }
+        public Expression[] Arguments { get; private set; }
```
marcovisserFurore: Is this needed? Because this is a breaking change: the signature of the property `Arguments` has changed

gstoian: Changed it back. It works as we discussed.

marcovisserFurore:
This solves https://github.com/FirelyTeam/fhir-net-api/issues/1128

PR #13, opened 2019-11-12
Detect circular reference

gstoian:
#465
Created BaseScopedNode and updated ScopedNode to inherit it.


marcovisserFurore (commented):
We are rejecting this PR, because in the new validator (2.1) this functionality should be built in.

PR #16, opened 2019-11-26
Support: faster EnumUtility.GetLiteral()

michelemottini:
Instead of using reflection every time use the same cache used when parsing - it is more than 10 times faster

EnumUtility.GetLiteral() is called for every assignment to a Code property of a POCO (eg Patient.Gender) - so its performance is material for any application that creates a lot of POCO objects

Fixes https://github.com/FirelyTeam/fhir-net-api/issues/1169

### DIFF
--- a/src/Hl7.Fhir.Support.Tests/Utility/EnumMappingTest.cs
+++ b/src/Hl7.Fhir.Support.Tests/Utility/EnumMappingTest.cs
@@ -9,7 +9,6 @@
 using System.Diagnostics;
 using Hl7.Fhir.Utility;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
-using Hl7.Fhir.Model;
 
 namespace Hl7.Fhir.Support.Tests.Utils
 {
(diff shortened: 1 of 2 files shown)


marcovisserFurore:
Also pulled for develop-1.x

michelemottini:
Thanks @marcovisserFurore

marcovisserFurore:
A big thank you to you @michelemottini . You did the commit ;)

michelemottini:
When would a new version of the Hl7.Fhir.Support package with this change be released?

marcovisserFurore:
We are working now on the technical correction (version 4.0.1 of the Fhir spec). When that is done, we will create a new release (1.5.0). My estimate will be before Christmas. 
You can follow our planning on this board: https://github.com/FirelyTeam/fhir-net-api/projects/13 
When the lanes 'Todo' and 'Progress' are empty we will create a release.

michelemottini:
Got it, thanks

PR #19, opened 2020-01-07
Bugfix/1192 Add IElementDefinitionSummary.DefaultTypeName

alexzautke:
Preparation to solve https://github.com/FirelyTeam/fhir-net-api/issues/1192


src/Hl7.Fhir.Support/Specification/ElementDefinitionSummary.cs
```diff
@@ -16,14 +16,15 @@ public class ElementDefinitionSummary : IElementDefinitionSummary
         private ElementDefinitionSummary() { }
 
         public ElementDefinitionSummary(string elementName, bool isCollection, bool isChoice,
```
marcovisserFurore: This is a public constructor on a public class. We cannot change this because we are going to affect the interface. Better is to create an overload of this constructor with an extra parameter

alexzautke: Added new overload of constructor

PR #20, opened 2020-01-07
Bugfix/1191 Handle XmlRepresentation.TypeAttr

alexzautke:
Fixes https://github.com/FirelyTeam/fhir-net-api/issues/1191
Unit test in separate PR for fhir-net-api.

### DIFF
--- a/src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
+++ b/src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
@@ -166,6 +166,21 @@ namespace Hl7.Fhir.ElementModel
                         raiseTypeError($"Choice element '{current.Name}' is suffixed with unexpected type '{suffix}'", current, location: current.Location);
                 }
             }
+            else if (info.Representation == XmlRepresentation.TypeAttr) // May be used by models other then FHIR, e.g. CCDA represented by a StructureDefinition
+            {
+                if(info.Type.Count() == 1)
+                {
+                    instanceType = info.Type.Single().GetTypeName();
+                }
+                else
+                {
+                    var typeName = current.Children("type").FirstOrDefault()?.Text;
+                    if (typeName != null)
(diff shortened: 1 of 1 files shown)


alexzautke:
Blocked until #19 is merged.

marcovisserFurore:
> Blocked until #19 is merged.

So that's why the build is now failing?

alexzautke:
Yes, but I didn't want to merge the PRs, so it's easier to review.

src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
```diff
@@ -166,6 +166,21 @@ private string deriveInstanceType(ISourceNode current, IElementDefinitionSummary
...
+            {
+                if(info.Type.Count() == 1)
+                {
+                    instanceType = info.Type.Single().GetTypeName();
+                }
+                else
+                {
+                    var typeName = current.Children("type").FirstOrDefault()?.Text;
+                    if (typeName != null)
+                        instanceType = info.Type.Where(type => type.GetTypeName().EndsWith(typeName)).FirstOrDefault().GetTypeName();
```
marcovisserFurore: Unfortunately this will fail. For example when the instance has the type `PQ` then the StructureDefinition has 3 possible candidates: `PQ`, `IVL_PQ` and `RTO_PQ_PQ` when we use the method `EndsWith()`. Incidentally it takes the first so we are safe, but it could wrong if the order is different.

marcovisserFurore: So you need something like this:

```
public static string Last(this string str) 
{
   var pos = str.LastIndexOf('/');
   return pos > -1 ? str.Substring(pos + 1) : str;
}

```
and then
```
instanceType = info.Type.Where(type => type.GetTypeName().Last() == typeName).FirstOrDefault().GetTypeName();
```

alexzautke: Changed like requested, 'Last' was a tricky name to choose as it's already in use.

PR #23, opened 2020-01-10
Bugfix/1193 Check SD.type when trying to map primitives in logical models

alexzautke:
This PR solves #1193, now using the correct branch (based on develop-1.x)

### DIFF
--- a/src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
+++ b/src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
@@ -174,6 +174,21 @@ namespace Hl7.Fhir.ElementModel
                         raiseTypeError($"Choice element '{current.Name}' is suffixed with unexpected type '{suffix}'", current, location: current.Location);
                 }
             }
+            else if (info.Representation == XmlRepresentation.TypeAttr) // May be used by models other then FHIR, e.g. CCDA represented by a StructureDefinition
+            {
+                if(info.Type.Count() == 1)
+                {
+                    instanceType = info.Type.Single().GetTypeName();
+                }
+                else
+                {
+                    var typeName = current.Children("type").FirstOrDefault()?.Text;
+                    if (typeName != null)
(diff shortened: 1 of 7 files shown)


src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
```diff
@@ -120,6 +120,14 @@ public object Value
                     return null;
                 }
 
+                // Type might still be a custom "primitive" type for logical models
+                if (Uri.IsWellFormedUriString(InstanceType, UriKind.Absolute))
+                {
+                    var summary = Provider.Provide(InstanceType);
+                    var valueType = summary?.GetElements().Where(e => e.ElementName.Equals("value")).FirstOrDefault()?.Type.FirstOrDefault()?.GetTypeName();
```
marcovisserFurore:
```suggestion
                    var valueType = summary?.GetElements().FirstOrDefault(e => e.ElementName.Equals("value"))?.Type.SingleOrDefault()?.GetTypeName();
```

marcovisserFurore: So what you are doing here is the following:
- when the InstanceType is an absolute Uri (for example `http://hl7.org/fhir/cda/StructureDefinition/CD`) then search for this StructureDefinition
- Then loop through the ElementDefintion of this type until you found the ElementDefintion with the name `value`.
- You will use this type of the value

I cannot find the value in for example `http://hl7.org/fhir/cda/StructureDefinition/CD`

alexzautke: I think it's an oversight in the StructureDefinitions for CDA. Following the current SDs there is no way for an implementation to know what they real underlying type of the CDA primitives are. Most datatypes have the value element present. While there is no guidance from the FHIR spec, we could try to discuss this topic on Zulip. For now I would be in favour of adding a "value" element to the StructureDefinitions on the primitive datatypes where it's missing. Even the FHIR core primitive types are defining their real datatypes, so the same should be true for logical primitive types.

PR #24, opened 2020-01-13
Feature/1208 xhtml in ccda

webersteffen:
IssueNumber:1208

With this feature the ccda tags with a xhtml type will be translated to one node in ITypedelement. They will be still in a tree structure after parsing to an ISourceNode type! The Logic is implemented, so that the normal FHIR procedures are not effected and the new logic only works with ccda/xhtml nodes.

### DIFF
--- a/src/Hl7.Fhir.ElementModel/IXHtml.cs
+++ b/src/Hl7.Fhir.ElementModel/IXHtml.cs
@@ -1,0 +1,7 @@
+﻿namespace Hl7.Fhir.ElementModel
+{
+    public interface IXHtml
+    {
+        string Serialize(ISourceNode node);
+    }
+}
--- a/src/Hl7.Fhir.ElementModel/IXHtml.cs
+++ b/src/Hl7.Fhir.ElementModel/IXHtml.cs
@@ -1,7 +1,0 @@
-﻿namespace Hl7.Fhir.ElementModel
-{
-    public interface IXHtml
-    {
-        string OriginalElement { get; }
-    }
-}
--- a/src/Hl7.Fhir.ElementModel/IXHtml.cs
+++ b/src/Hl7.Fhir.ElementModel/IXHtml.cs
@@ -2,6 +2,6 @@
 {
     public interface IXHtml
     {
-        string Serialize(ISourceNode node);
+        string OriginalElement { get; }
     }
 }
--- a/src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
+++ b/src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
@@ -166,6 +166,21 @@ namespace Hl7.Fhir.ElementModel
                         raiseTypeError($"Choice element '{current.Name}' is suffixed with unexpected type '{suffix}'", current, location: current.Location);
                 }
             }
+            else if (info.Representation == XmlRepresentation.TypeAttr) // May be used by models other then FHIR, e.g. CCDA represented by a StructureDefinition
+            {
+                if(info.Type.Count() == 1)
+                {
+                    instanceType = info.Type.Single().GetTypeName();
+                }
+                else
+                {
+                    var typeName = current.Children("type").FirstOrDefault()?.Text;
+                    if (typeName != null)
+                        instanceType = info.Type.Where(type => type.GetTypeName().EndsWith(typeName)).FirstOrDefault().GetTypeName();
+                    else
+                        instanceType = info.DefaultTypeName;
+                }
+            }
             else
             {
                 var tp = info.Type.Single();
--- a/src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
+++ b/src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
@@ -187,7 +187,7 @@ namespace Hl7.Fhir.ElementModel
                 {
                     var typeName = current.Children("type").FirstOrDefault()?.Text;
                     if (typeName != null)
-                        instanceType = info.Type.Where(type => type.GetTypeName().EndsWith(typeName)).FirstOrDefault().GetTypeName();
+                        instanceType = info.Type.Where(type => typeFromLogicalModelCanonical(type).Equals(typeName)).FirstOrDefault().GetTypeName();
                     else
                         instanceType = info.DefaultTypeName;
                 }
@@ -199,6 +199,13 @@ namespace Hl7.Fhir.ElementModel
             }
 
             return instanceType;
+        }
+
+        private string typeFromLogicalModelCanonical(this ITypeSerializationInfo info)
+        {
+            var type = info.GetTypeName();
+            var pos = type.LastIndexOf('/'); // Match the "raw" type name from the complete type name of the logical model type (absolute url) 
+            return pos > -1 ? type.Substring(pos + 1) : type;
         }
 
         private bool tryGetBySuffixedName(Dictionary<string, IElementDefinitionSummary> dis, string name, out IElementDefinitionSummary info)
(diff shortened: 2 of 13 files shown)


src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
```diff
@@ -18,6 +18,9 @@ namespace Hl7.Fhir.ElementModel
 {
     internal class TypedElementOnSourceNode : ITypedElement, IAnnotated, IExceptionSource, IShortPathGenerator
     {
+        private readonly string xHtmlInstancetype = "xhtml";
```
marcovisserFurore: we normally use the style `private const string XHTML_INSTANCETYPE = "xHtml"`

src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
```diff
@@ -18,6 +18,9 @@ namespace Hl7.Fhir.ElementModel
...
+        private readonly string htmlDivTagName = "div";
```
marcovisserFurore: See above

src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
```diff
@@ -259,12 +281,22 @@ private IEnumerable<TypedElementOnSourceNode> enumerateElements(Dictionary<strin
...
         }
 
         public IEnumerable<ITypedElement> Children(string name = null)
         {
+            //If we have a xhtml typed node and there was not an div tag around the content
+            //the won´t look at the children of this node, since there will be no types
+            //matching the html tags.
+            if (this.InstanceType == xHtmlInstancetype && Name != htmlDivTagName)
+            {
+                return new ITypedElement[] { };
```
marcovisserFurore: Use `return Enumerable.Empty<ITypedElement>();`

src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
```diff
@@ -247,6 +251,24 @@ private IEnumerable<TypedElementOnSourceNode> enumerateElements(Dictionary<strin
...
+                {
+                    string xml = string.Empty;
+                    foreach (var xmlChild in scan.Children())
+                    {
+                        var ix = xmlChild.Annotation<IXHtml>();
+                        xml += ix?.OriginalElement;
+
+                    }
+                    isCcdaXhtmlNode = true;
+                    source = SourceNode.Valued(scan.Name, xml);
```
marcovisserFurore: Try to yield here already. To do this, then move the prettypath logic before this block

src/Hl7.Fhir.ElementModel/IXHtml.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Hl7.Fhir.ElementModel
+{
+    public interface IXHtml
```
marcovisserFurore: We need another name here. This was just an example.

src/Hl7.Fhir.ElementModel/IXHtml.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Hl7.Fhir.ElementModel
```
marcovisserFurore: Add the following in the header:
```
/* 
 * Copyright (c) 2020, Firely (info@fire.ly) and contributors
 * See the file CONTRIBUTORS for details.
 * 
 * This file is licensed under the BSD 3-Clause license
 * available at https://github.com/FirelyTeam/fhir-net-api/blob/master/LICENSE
 */
```

src/Hl7.Fhir.Serialization/FhirXmlNode.cs
```diff
@@ -437,5 +437,38 @@ object checkRepresentation(ITypedElement node, IExceptionSource ies, object _)
...
+        public string OriginalElement
+        {
+            get {
+                if (((XElement)Current.Parent).Name.Namespace != ((XElement)Current).Name.Namespace)
+                {
+                    return ((XElement)Current).ToString(SaveOptions.DisableFormatting);
+                }
+                else
+                {
+                    var test = StripNamespaces((XElement)Current);
```
marcovisserFurore: not really a nice variable name `test`

src/Hl7.Fhir.Serialization/FhirXmlNode.cs
```diff
@@ -437,5 +437,38 @@ object checkRepresentation(ITypedElement node, IExceptionSource ies, object _)
 
         private ExceptionNotification buildException(string message) => ExceptionNotification.Error(
                 new StructuralTypeException("Parser: " + message));
+
+        public string OriginalElement
+        {
+            get {
+                if (((XElement)Current.Parent).Name.Namespace != ((XElement)Current).Name.Namespace)
```
marcovisserFurore: create a local variable like this `var current = Current as XElement` and use this in this getter (instead of casting it on 3 points)

src/Hl7.Fhir.ElementModel/IXHtml.cs
marcovisserFurore: Perhaps this will cover it: `ICcdaInfoSupplier`

src/Hl7.Fhir.ElementModel/IXHtml.cs
```diff
@@ -0,0 +1,7 @@
...
+{
+    public interface IXHtml
+    {
+        string OriginalElement { get; }
```
marcovisserFurore: Also this was just a blabla name. I think we should call it `XHtmlText`

src/Hl7.Fhir.ElementModel/IXHtml.cs
webersteffen: added

src/Hl7.Fhir.ElementModel/IXHtml.cs
webersteffen: Sounds good to me. Renamed it.

src/Hl7.Fhir.ElementModel/IXHtml.cs
webersteffen: Renamed it.

src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
webersteffen: Of course. I changed it.

src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
webersteffen: Changed this too.

src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
webersteffen: Did this while adding a continue to make this double yield working. Also moved prettypath logic before and removed isCcdaXhtmlNode since its not necessary any more.

src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
webersteffen: Changed it.

src/Hl7.Fhir.Serialization/FhirXmlNode.cs
webersteffen: Done.

src/Hl7.Fhir.Serialization/FhirXmlNode.cs
webersteffen: I should start naming things seriously even when I testing around. Chagend this to: stripedXElement

PR #27, opened 2020-02-11
CCDA support

marcovisserFurore:


src/Hl7.Fhir.ElementModel/ICcdaInfoSupplier.cs
```diff
@@ -0,0 +1,15 @@
...
+ * Copyright (c) 2020, Firely (info@fire.ly) and contributors
+ * See the file CONTRIBUTORS for details.
+ * 
+ * This file is licensed under the BSD 3-Clause license
+ * available at https://github.com/FirelyTeam/fhir-net-api/blob/master/LICENSE
+ */
+
+namespace Hl7.Fhir.ElementModel
+{
+    public interface ICcdaInfoSupplier
```
marcovisserFurore: Should be ICdaInfoSupplier

src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
```diff
@@ -166,6 +177,21 @@ private string deriveInstanceType(ISourceNode current, IElementDefinitionSummary
...
+            {
+                if (info.Type.Count() == 1)
+                {
+                    instanceType = info.Type.Single().GetTypeName();
+                }
+                else
+                {
+                    var typeName = current.Children("type").FirstOrDefault()?.Text;
+                    if (typeName != null)
+                        instanceType = info.Type.Where(type => typeFromLogicalModelCanonical(type).Equals(typeName)).FirstOrDefault().GetTypeName();
```
mmsmits: Moet hier niet nog een null check, achter FirstOrDefault?

src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
```diff
@@ -248,12 +249,23 @@ private IEnumerable<TypedElementOnSourceNode> enumerateElements(Dictionary<strin
...
+                //If we encounter a xhtml node in a ccda document we will flatten all childnodes
+                //and use there content to build up the xml.
+                //The xml will be put in this node and children will be ignored.
+                if (instanceType == "xhtml" && info.Representation == XmlRepresentation.CdaText)
                 {
-                    var sn = SourceNode.FromNode(scan, false, new[] { typeof(IXHtml) });
-                    var ix = sn.Annotation<IXHtml>();
-                    var xml = ix?.Serialize(scan);
+                    string xml = string.Empty;
+                    foreach (var xmlChild in scan.Children())
```
ewoutkramer: Shouldn't combining several text nodes into one be a responsibility for the underlying SourceNode?  I agree we have the type information here - so that could be a reason to only handle this here, but could this possibly be generalized so we can push it down?  Then we don't need to create another node.  This creates additional GC pressure (there are *many* such nodes in a CCDA document, and we're potentially doubling our memory footprint).

src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
```diff
@@ -267,12 +279,22 @@ private IEnumerable<TypedElementOnSourceNode> enumerateElements(Dictionary<strin
...
+                yield return new TypedElementOnSourceNode(this, source, info, instanceType, prettyPath);
             }
         }
 
         public IEnumerable<ITypedElement> Children(string name = null)
         {
+            //If we have a xhtml typed node and there was not an div tag around the content
+            //the won´t look at the children of this node, since there will be no types
+            //matching the html tags.
+            if (this.InstanceType == "xhtml" && Name != "div")
```
ewoutkramer: This feels like duplicate code, since in FHIR we already had xhtml + div nodes too, and so this case is already handled. Possibly in enumerateElements() (sorry, cannot double check from github web interface where it is exactly).

src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
```diff
@@ -267,12 +279,22 @@ private IEnumerable<TypedElementOnSourceNode> enumerateElements(Dictionary<strin
...
+            {
+                return new ITypedElement[] { };
```
ewoutkramer: I prefer Enumerable.Empty<ITypedElement>(), but that might be stylistic nitpicking.

src/Hl7.Fhir.Serialization/FhirXmlNode.cs
```diff
@@ -438,10 +438,7 @@ object checkRepresentation(ITypedElement node, IExceptionSource ies, object _)
         private ExceptionNotification buildException(string message) => ExceptionNotification.Error(
                 new StructuralTypeException("Parser: " + message));
 
-        public string Serialize(ISourceNode node)
-        {
-            return node.ToXml();
-        }
+        public string OriginalElement => ((XElement)Current).ToString(SaveOptions.DisableFormatting);
```
ewoutkramer: SourceNode already provides this kind of access to the underlying data, could we not push this down into sourcenode, so it ends up with all the other "source xml info" code?

src/Hl7.Fhir.Serialization/FhirXmlNode.cs
```diff
@@ -438,7 +438,37 @@ object checkRepresentation(ITypedElement node, IExceptionSource ies, object _)
         private ExceptionNotification buildException(string message) => ExceptionNotification.Error(
                 new StructuralTypeException("Parser: " + message));
 
-        public string OriginalElement => ((XElement)Current).ToString(SaveOptions.DisableFormatting);
+        public string OriginalElement
+        {
+            get {
+                if (((XElement)Current.Parent).Name.Namespace != ((XElement)Current).Name.Namespace)
```
ewoutkramer: Null reference exception when you call this property on the root of a document, since there won't be a .Parent.

ewoutkramer: Feels a bit like a hack?  If it is, we better add a bit of documentation here.

src/Hl7.Fhir.Serialization/FhirXmlNode.cs
```diff
@@ -438,7 +438,37 @@ object checkRepresentation(ITypedElement node, IExceptionSource ies, object _)
...
+                else
+                {
+                    var test = StripNamespaces((XElement)Current);
+                    return (test.ToString(SaveOptions.DisableFormatting));
+                }
+            }
+            
+        }
+
+        private XElement StripNamespaces(XElement rootElement)
```
ewoutkramer: I have no idea what and why this is, can we add documentation?

ewoutkramer: It feels like this property is expensive with all the namespace manipulation. We should cache the result I think.

src/Hl7.Fhir.Serialization/FhirXmlNode.cs
```diff
@@ -17,9 +17,9 @@
 
 namespace Hl7.Fhir.Serialization
 {
-    public partial class FhirXmlNode : ISourceNode, IResourceTypeSupplier, IAnnotated, IExceptionSource
-    {  
-        internal FhirXmlNode(XObject node, FhirXmlParsingSettings settings) 
+    public partial class FhirXmlNode : ISourceNode, IResourceTypeSupplier, IAnnotated, IExceptionSource, IXHtml
```
ewoutkramer: Yes! This is the right place - sorry I am reviewing commits in historical order, so I commented on IXHtml being implemented by TypedElementOnSourceNode. This here seems to be the correct solution.

src/Hl7.Fhir.ElementModel/ICdaInfoSupplier.cs
```diff
@@ -8,8 +8,11 @@
 
 namespace Hl7.Fhir.ElementModel
 {
-    public interface ICcdaInfoSupplier
+    public interface ICdaInfoSupplier
```
ewoutkramer: Or Hl7v3InfoSupplier?  I don't even think this is CDA specific?  Well, the use of XHTML may be, but there might be other information (like information about the original xsi:type attribute in the source data) that you make available here, and this is more general for the v3 serialization, and not specific to CDA.

src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
ewoutkramer: It should be, but the SourceNode does not know it is dealing with narrative here - so it can't. Maybe the higher-level ITypedElement should indeed be implementing this, but currently, it does not work for mixed-mode narrative. See .NET API issue 1274.

src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
```diff
@@ -120,6 +120,14 @@ public object Value
                     return null;
                 }
 
+                // Type might still be a custom "primitive" type for logical models
+                if (Uri.IsWellFormedUriString(InstanceType, UriKind.Absolute))
+                {
+                    var summary = Provider.Provide(InstanceType);
+                    var valueType = summary?.GetElements().Where(e => e.ElementName.Equals("value")).FirstOrDefault()?.Type.FirstOrDefault()?.GetTypeName();
```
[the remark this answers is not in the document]
ewoutkramer: See .NET API issue 1276.

src/Hl7.Fhir.ElementModel/ICdaInfoSupplier.cs
ewoutkramer: It's not CDA specific - we want more general "give me the original wire-representation of this node" functionality here.  This is connected to how SourceNode and TypedElement cooperate on nodes that should actually not be parsed.  SourceNode has no type information present to deal with that, but TypedElement is too late to recombine these nodes into their original representation. See also .NET API issue 1274.

src/Hl7.Fhir.Serialization/FhirXmlNode.cs
ewoutkramer: I don't agree we should remove the namespace - but it's too late to change this now. Will mark all of this functionality as [Obsolete] so we can review this in the next version of the API.

src/Hl7.Fhir.Serialization/XObjectFhirXmlExtensions.cs
```diff
@@ -79,9 +79,10 @@ public static string GetValue(this XObject current)
         {
             if (current.AtXhtmlDiv())
                 return ((XElement)current).ToString(SaveOptions.DisableFormatting);
-            else
-                return current is XElement xelem ?
-                    xelem.Attribute("value")?.Value.Trim() : current.Value();
+            if (current.FirstChild() is XText) return current.Value();
+            return current is XElement xelem ?
```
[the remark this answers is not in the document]
ewoutkramer: Yes.

PR #33, opened 2020-04-14
#1312 Make resolvers & snapgen work asynchronously

ewoutkramer:
Add infrastructure to create asynchronous resolvers, and update the snapshot generator to use them.

Note: API still has many places where it uses the now-obsolete synchronous calls.

### DIFF
--- a/src/Hl7.Fhir.Support.Poco/Specification/Source/CachedResolver.cs
+++ b/src/Hl7.Fhir.Support.Poco/Specification/Source/CachedResolver.cs
@@ -43,7 +43,7 @@ namespace Hl7.Fhir.Specification.Source
         {
             Source = source as IResourceResolver;
 #pragma warning restore CS0618 // Type or member is obsolete
-            AsyncSource = source.AsAsync();
+            AsyncResolver = source.AsAsync();
             CacheDuration = cacheDuration;
 
             _resourcesByUri = new Cache<Resource>(id => InternalResolveByUri(id), CacheDuration);
@@ -51,13 +51,16 @@ namespace Hl7.Fhir.Specification.Source
         }
 
         /// <summary>
-        /// Returns the child source, as specified in the constructor.
+        /// Returns the resolver for which access is cached, as specified in the constructor.
         /// </summary>
         /// <remarks>May return <code>null</code> if the source is an exclusively asynchronous resolver.</remarks>
         [Obsolete("CachedResolver now works best with asynchronous resolvers. Use the AsyncSource property instead.")]
         public IResourceResolver Source { get; private set; }
 
-        public IAsyncResourceResolver AsyncSource { get; }
+        /// <summary>
+        /// Returns the resolver for which access is cached, as specified in the constructor.
+        /// </summary>
+        public IAsyncResourceResolver AsyncResolver { get; }
 
         /// <summary>Gets the default expiration time of a cache entry.</summary>
         public int CacheDuration { get; }
@@ -182,14 +185,14 @@ namespace Hl7.Fhir.Specification.Source
 
         internal async Task<Resource> InternalResolveByUri(string url)
         {
-            var resource = await AsyncSource.ResolveByUriAsync(url);
+            var resource = await AsyncResolver.ResolveByUriAsync(url);
             OnLoad(url, resource);
             return resource;
         }
 
         internal async Task<Resource> InternalResolveByCanonicalUri(string url)
         {
-            var resource = await AsyncSource.ResolveByCanonicalUriAsync(url);
+            var resource = await AsyncResolver.ResolveByCanonicalUriAsync(url);
             OnLoad(url, resource);
             return resource;
         }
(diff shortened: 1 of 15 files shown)


src/Hl7.Fhir.Support/Utility/DebuggerDisplayExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Hl7.Fhir.Utility
+{
+    public static class DebuggerDisplayExtensions
+    {
+        public const string DEBUGGER_DISPLAY_PROP_NAME = "DebuggerDisplay";
+        public static string DebuggerDisplayString(this object target)
+        {
+            var debuggerDisplay = ReflectionHelper.FindProperty(target.GetType(), DEBUGGER_DISPLAY_PROP_NAME);
+            if (debuggerDisplay == null) return null;
```
marcovisserFurore, line 17:
```suggestion
            if (debuggerDisplay is null) return null;
```

src/Hl7.Fhir.Support.Poco/Specification/Source/IResourceResolver.cs
```diff
@@ -0,0 +1,53 @@
...
+
+using System;
+using System.Threading.Tasks;
+using Hl7.Fhir.Model;
+
+namespace Hl7.Fhir.Specification.Source
+{
+    /// <summary>Interface for resolving FHIR artifacts by (canonical) uri.</summary>
+#pragma warning disable CS0618 // Type or member is obsolete
+    public interface IResourceResolver : ISyncOrAsyncResourceResolver
```
marcovisserFurore, line 17: Should we not mark this interface as Obsolete?

src/Hl7.Fhir.Support.Poco/Specification/Source/CachedResolver.cs
```diff
@@ -0,0 +1,307 @@
...
+        /// <summary>Occurs when a artifact is loaded into the cache.</summary>
+        public event LoadResourceEventHandler Load;
+
+        /// <summary>Called when an artifact is loaded into the cache.</summary>
+        protected virtual void OnLoad(string url, Resource resource) => Load?.Invoke(this, new LoadResourceEventArgs(url, resource));
+
+        internal async Task<Resource> InternalResolveByUri(string url)
+        {
+            var resource = await AsyncResolver.ResolveByUriAsync(url);
+            OnLoad(url, resource);
```
marcovisserFurore, line 189: Should this also not be async?

src/Hl7.Fhir.Support.Poco/Specification/Source/IResourceResolver.cs
ewoutkramer: That is a difficult question that I have spent quite some time on. To be perfectly honest, in v2.x, the synchronous interface will still be used in a large part of the codebase. So although we plan to go all-async, this is certainly not yet the case. With this in mind, I decided to NOT make it obsolete yet - since it is in actual use, without any fixed plans to remove it. That's also why I went out of my way to support both sync and async scenarios.

Another very practical argument, that is a consequence of this: Marking this obsolete will generate so many obsolete messages that they become totally useless - you'll just skip looking at them.

src/Hl7.Fhir.Support.Poco/Specification/Source/CachedResolver.cs
ewoutkramer: We should make this a private method - then we don't have to put the Async suffix.

marcovisserFurore: I meant the OnLoad. So `await OnLoad(url, resource);`. 

So, should we make OnLoad async?

src/Hl7.Fhir.Support.Poco/Specification/Source/IResourceResolver.cs
ewoutkramer: Maybe `[Obsolete]` on constructor - at least we discourage new use of the old interface?

src/Hl7.Fhir.Support.Poco/Specification/Source/CachedResolver.cs
ewoutkramer: Note that the I/O has already been done, this is supposed to be just a quick signal to interested parties that I/O has finished, so no async necessary here.  (and I didn't want to go into async events yet).

src/Hl7.Fhir.Support.Poco/Specification/Source/IResourceResolver.cs
ewoutkramer: Actually, this is not possible, sorry ;-)   There is only one constructor on e.g. DirectorySource, and it takes ISyncOrAsyncResolver, so it cannot be marked `[Obsolete]`.

marcovisserFurore: Ok, then we leave it like this. Thanks for the research.

PR #34, opened 2020-04-16
fix: 1140 casting did not account for empty primitives

ewoutkramer:
This fixes the FhirPath error on the STU3 eld13 expression.


marcovisserFurore:
Fixes FirelyTeam/fhir-net-api#1140

PR #38, opened 2020-04-20
Factory pattern for assertions

abelenthoven:


src/Hl7.Fhir.Validation/Schema/IElementSchema.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    public interface IElementSchema: IAssertion
+    {
+        Uri Id { get; }
+
+        Assertions Members { get; }
+
+        IElementSchema With(IEnumerable<IAssertion> additional);
+
+        Assertions Validate(IEnumerable<ITypedElement> input, ValidationContext vc);
```
marcovisserFurore, line 18: This is part of `IGroupValidatable`.  So I think `IElementSchema` should derive from `IAssertion` and `IGroupValidatable`

src/Hl7.Fhir.Validation/Schema/ElementSchema.cs
```diff
@@ -15,11 +15,11 @@
 
 namespace Hl7.Fhir.Validation.Schema
 {
-    public class ElementSchema : IAssertion, IMergeable, IGroupValidatable
+    public class ElementSchema : IElementSchema, IMergeable, IGroupValidatable
```
marcovisserFurore, line 18: When `IElementSchema` is derived from `IGroupValidatable` then `IGroupValidatable`  can be removed here

PR #40, opened 2020-04-23
fix: 1304 - also raise an exception if the ResourceType is empty

marcovisserFurore:
Fixes FirelyTeam/fhir-net-api#1304


ewoutkramer:
Isn't this problem present in the XML parser too?

marcovisserFurore:
I don't think so, because the element `resourceType` is only available in Json. In Xml you will have this:
```
<Patient xmlns='http://hl7.org/fhir'>
  <id value="123"/>
</Patient>
```

PR #44, opened 2020-05-07
#1356: Move all primitives to common

ewoutkramer:
An earlier PR moved the primitives necessary to represent Resource to common, this PR finishes the work and moves all other primitive FHIR type POCO's to common.


ewoutkramer:
Fixes FirelyTeam/fhir-net-api#1356.

PR #50, opened 2020-05-24
Implemented Fhir Patch

anyname2:
Added support for FHIR Patch operations. Only FHIRPath patch operation is implemented.
Addresses: https://github.com/FirelyTeam/fhir-net-api/issues/1374


src/Hl7.Fhir.Patch/Adapters/IObjectAdapter.cs
```diff
@@ -0,0 +1,122 @@
...
+ */
+
+using Hl7.Fhir.Patch.Operations;
+
+namespace Hl7.Fhir.Patch.Adapters
+{
+    /// <summary>
+    /// Defines the operations that can be performed on a patch document.
+    /// </summary>
+    public interface IObjectAdapter
```
cknaap, line 16: I find the distinction between (I)ObjectAdapter and IAdapter confusing.

src/Hl7.Fhir.Patch/IPatchDocument.cs
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+using Hl7.Fhir.Patch.Operations;
+using System.Collections.Generic;
+
+namespace Hl7.Fhir.Patch
+{
+    public interface IPatchDocument
+    {
+        IList<Operation> GetOperations();
```
cknaap, line 16: Neither this method nor its implementation in `PatchDocument` is used. Instead only `PatchDocument.ApplyTo(ElementNode)` is used. 
Either remove the interface altogether or adjust it to the actually used public methods.

src/Hl7.Fhir.Patch/Internal/CollectionAdapter.cs
```diff
@@ -0,0 +1,149 @@
...
+    /// <summary>
+    /// This API supports infrastructure and is not intended to be used
+    /// directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    internal class CollectionAdapter : IAdapter
+    {
+        public virtual bool TryAdd(
+            object target,
+            string name,
+            IStructureDefinitionSummaryProvider contractResolver,
```
cknaap, line 26: Maybe provide the `IStructureDefinitionSummaryProvider` just once in the constructor?

src/Hl7.Fhir.Patch/Internal/ElementNodeAdapter.cs
```diff
@@ -0,0 +1,174 @@
...
+    /// <summary>
+    /// This API supports infrastructure and is not intended to be used
+    /// directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    internal class ElementNodeAdapter : IAdapter
+    {
+        public virtual bool TryAdd(
+            object target,
+            string name,
+            IStructureDefinitionSummaryProvider contractResolver,
```
cknaap, line 26: Maybe provide the `IStructureDefinitionSummaryProvider` just once in the constructor?

src/Hl7.Fhir.Patch/Internal/ElementNodeAdapter.cs
```diff
@@ -0,0 +1,174 @@
...
+                    return true;
+                }
+            }
+
+            PropertyValue = Enumerable.Empty<ITypedElement>();
+            PropertyDefinition = null;
+            return false;
+        }
+
+        protected virtual bool TryConvertValue (object value, IElementDefinitionSummary propertyDefinition, out ElementNode convertedValue)
```
cknaap, line 156: Naming: More of a 'choice check' than an actual conversion.

src/Hl7.Fhir.Patch/Internal/ElementNodeAdapter.cs
```diff
@@ -0,0 +1,174 @@
...
+            IStructureDefinitionSummaryProvider contractResolver,
+            out object value,
+            out string errorMessage)
+        {
+            value = target;
+            errorMessage = null;
+            return true;
+        }
+
+        protected virtual bool TryGetProperty(
```
cknaap, line 134: Naming: In general we use 'Element' when working with ITypedElement and 'Property' when working with POCO's.

src/Hl7.Fhir.Patch/Internal/ElementNodeAdapter.cs
cknaap: This method is duplicated in `CollectionAdapter` - consider a helper method.

src/Hl7.Fhir.Patch/Internal/IAdapter.cs
```diff
@@ -0,0 +1,50 @@
...
+
+using Hl7.Fhir.Specification;
+
+namespace Hl7.Fhir.Patch.Internal
+{
+    /// <summary>
+    /// This API supports infrastructure and is not intended to be used
+    /// directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    public interface IAdapter
```
cknaap, line 17: Naming: `Adapter` suggests the use of the Adapter pattern, which is not the case here. Something with Editor, Updater or maybe even Patcher in it might work better.

src/Hl7.Fhir.ElementModel/ElementNode.cs
```diff
@@ -94,6 +94,37 @@ public ElementNode Add(IStructureDefinitionSummaryProvider provider, string name
             return Add(provider, child);
         }
 
+        public ElementNode Insert (IStructureDefinitionSummaryProvider provider, ElementNode child, int index, string name = null)
```
cknaap, line 97: To avoid confusion I would keep only this overload, and without the `name` parameter (caller should set the correct name in `child`).

src/Hl7.Fhir.ElementModel/ElementNode.cs
```diff
@@ -94,6 +94,37 @@ public ElementNode Add(IStructureDefinitionSummaryProvider provider, string name
...
+                throw new ArgumentNullException(nameof(child));
+
+            var childName = name ?? child.Name;
+            if ( child.Name == null ) throw Error.Argument($"The ElementNode given should have its Name property set or the '{nameof(name)}' parameter should be given.");
+
+            var childIndices = ChildList.Select((x, i) => (child: x, index: (int?)i)).Where(kvp => kvp.child.Name == childName).Select(kvp => kvp.index).ToArray();
+            if(index < 0 || index > childIndices.Length) throw Error.Argument($"The index given is out of bounds of the array size.");
+
+            var insertAtIndex = childIndices.ElementAtOrDefault(index);
+            importChild(provider, child, name, insertAtIndex ?? childIndices.Length);
```
cknaap, line 111: Explanation on `importChild` states: "Will update the child to reflect it being a child of this element, but will not yet add the child at any position within this element."
So when *is* it positioned within this element then?

src/Hl7.Fhir.Patch/Adapters/IObjectAdapter.cs
anyname2: Both interfaces implement Adapter pattern and add methods to perform each of the Patch Operations to `ElementNode` and `IEnumerable<ITypedElement>`. 

`ObjectAdapter` is an abstraction over the concrete implemetations for each of the types supported by Patch Operatons (`ElementNodeAdapter` and `CollectionAdapter` respectively). 

`IObjectAdapter` and `IAdapter` are esseantially the same interface, but they needed to be different to avoid duplicating code. Some of the patch operations is just a combination of others (e.g. Move == Delete + Insert). `IObjectAdapter` is a public interface, which defines methods for each of the patch operations, while `IAdapter` is an internal interface that only defines the base methods that are necessary to implement all the other patch operations.

src/Hl7.Fhir.ElementModel/ElementNode.cs
anyname2: I have not made any changes to `importChild` method, but that comment on the method made me confused too

src/Hl7.Fhir.Patch/ElementNodePatchExtensions.cs
```diff
@@ -0,0 +1,113 @@
...
+namespace Hl7.Fhir.Patch
+{
+    public static class ElementNodePatchExtensions
+    {
+        /// <summary>
+        /// Apply <paramref name="patchDocument"/> to this <see cref="ElementNode"/>
+        /// </summary>
+        /// <param name="this"><see cref="ElementNode"/> to apply patch to</param>
+        /// <param name="patchDocument"><see cref="PatchDocument"/> to apply to this <see cref="ElementNode"/></param>
+        public static void Apply (this ElementNode @this, PatchDocument patchDocument)
```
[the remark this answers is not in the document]
anyname2: `@this` makes sense for me in extension methods. 
The intent here is to add a method to an exisitng class. If the method was added directly to a class we would use `this` keyword to refer to it.
But ok, I'll rename.

src/Hl7.Fhir.Patch/ElementNodePatchExtensions.cs
```diff
@@ -0,0 +1,113 @@
...
+ * available at https://github.com/FirelyTeam/fhir-net-api/blob/master/LICENSE
+ */
+
+using System;
+using Hl7.Fhir.ElementModel;
+using Hl7.Fhir.Patch.Adapters;
+
+namespace Hl7.Fhir.Patch
+{
+    public static class ElementNodePatchExtensions
```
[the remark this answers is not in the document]
anyname2: My thinking here was:

- If a user wants to chain multiple patch documents, the `ElementNode` overload will prevent making unnecessary copies
- If a user already has a mutable object, it makes sense to allow mutating it, rather than making a copy

But I agree, if there is no need to support the 2 use cases above, I will make the `ElementNode` overloads internal.

src/Hl7.Fhir.Patch/Operations/Operation.cs
```diff
@@ -0,0 +1,126 @@
...
+ * available at https://github.com/FirelyTeam/fhir-net-api/blob/master/LICENSE
+ */
+
+using System;
+using Hl7.Fhir.Patch.Adapters;
+using Hl7.FhirPath;
+
+namespace Hl7.Fhir.Patch.Operations
+{
+    public class Operation : OperationBase
```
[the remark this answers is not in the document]
anyname2: Refactored: now there is `Operation` and `OperationWithValue`

anyname2:
I have addressed most of the comments and refactored `Operation` class. Now there is a separate class for each type of operation.

I have considered moving implementation of individual operations from `ObjectAdapter` to their respective `Operation` classes, but that requires `IStructureDefinitionSummaryProvider`, `IAdapterFactory` and `Action<PatchError>`, which should be common to all operations within `PatchDocument` to be accessible from `Operation.Apply(...)`.
I am not sure what would be the best way to achieve this. @cknaap?

I still need to add more tests and look at `FhirPatch` POCO refactoring.

src/Hl7.Fhir.ElementModel/ElementNode.cs
marcovisserFurore: I agree

src/Hl7.Fhir.Patch/PatchDocument.cs
```diff
@@ -0,0 +1,128 @@
...
+using Hl7.Fhir.Specification;
+
+namespace Hl7.Fhir.Patch
+{
+    // Implementation details: the purpose of this type of patch document is to allow creation of such
+    // documents for cases where there's no class/DTO to work on. Typical use case: backend not built in
+    // .NET or architecture doesn't contain a shared DTO layer.
+    public class PatchDocument : IPatchDocument
+    {
+        public List<Operation> Operations { get; }
```
marcovisserFurore, line 25: Perhaps making there fields private? Maintaining this list is done by the method `Add`

anyname2:
> 
> 
> Because @cknaap is enjoying his holiday, I take over the review.
> 
> Thank you for this well documented and readable pull request.
> 
> I have a few small recommendations. If you want, you can change them. Otherwise we pull the request and we change some minor things ourselves. Let me know what you want to do.

I am going on holiday myself soon. I am happy you make the changes yourself.

marcovisserFurore:
We've decided to implement the `Patch` operation in another way. Therefor I close this PR.

PR #55, opened 2020-06-05
implement join function for fhirpath

webersteffen:
-added FpJoin to collectionoperators
-added join function to fp symboltable
-added two test for FpJoin

resolves FirelyTeam/fhir-net-api#1370

### DIFF
--- a/src/Hl7.FhirPath.Tests/Tests/BasicFunctionTests.cs
+++ b/src/Hl7.FhirPath.Tests/Tests/BasicFunctionTests.cs
@@ -12,7 +12,9 @@
 using Hl7.Fhir.ElementModel;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System.Collections.Generic;
+using Hl7.FhirPath.Functions;
 using System.Linq;
+using System;
 
 namespace Hl7.FhirPath.Tests
 {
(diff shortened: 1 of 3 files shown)


src/Hl7.FhirPath/FhirPath/Functions/CollectionOperators.cs
```diff
@@ -87,5 +87,22 @@ public static IEnumerable<ITypedElement> Navigate(this ITypedElement element, st
...
+        public static string FpJoin(this IEnumerable<ITypedElement> collection, string separator)
+        {
+            //if the collection is empty return the empty result
+            if (!collection.Any())
+                return string.Empty;
+
+            //only join collections with string values inside
+            foreach (var typedElement in collection)
+            {
+                if (typedElement.InstanceType != "string")
```
webersteffen, line 100: @marcovisserFurore : Hi marco. At this place I didnt found any good solution to check this type better than that. In Primitive.cs the types get set hard coded as string. Maybe there is some enum or somethig that I didn´t see?

src/Hl7.FhirPath/FhirPath/Functions/CollectionOperators.cs
```diff
@@ -112,5 +112,22 @@ public static IEnumerable<ITypedElement> Navigate(this ITypedElement element, st
...
```
marcovisserFurore, line 125: I would replace the foreach loop with this:

```
 if (!collection.All(c => c.Value is string))   
   throw Error.InvalidOperation("Join function can only be performed on string collections.");
```

src/Hl7.FhirPath.Tests/Tests/BasicFunctionTests.cs
```diff
@@ -74,6 +76,32 @@ public void TestStringSplit()
...
+        [TestMethod]
+        public void TestStringJoin()
+        {
+            var dummy = ElementNode.CreateList("This ", "is ", "one ", "sentence", ".");
+            var result = dummy.FpJoin(string.Empty);
+            Assert.IsNotNull(result);
+            Assert.AreEqual("This is one sentence.", result);
+
+            dummy = ElementNode.CreateList();
+            result = dummy.FpJoin("");
```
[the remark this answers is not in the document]
webersteffen: changed it

src/Hl7.FhirPath/FhirPath/Functions/CollectionOperators.cs
webersteffen: Thanks for the suggestion. I talked to ewout and replaced the impleentation with your suggestion.

PR #69, opened 2020-07-23
Introduce ICqlConvertible

ewoutkramer:
Introduction of ICqlConvertible, implemented by the Any types allows us to ask the Any values itself to convert to any of the other Any types, instead of putting this logic in the FhirPath engine.


src/Hl7.FhirPath/FhirPath/Functions/ConversionOperators.cs
```diff
@@ -19,129 +22,48 @@ internal static class ConversionOperators
...
-                    case "no":
-                    case "n":
-                    case "0":
-                    case "0.0":
-                        return false;
-                    default:
-                        return null;
-                }
-            }
+            if (!(focus is ICqlConvertible c)) return null;
```
marcovisserFurore, line 27: Maybe using an inline if?

PR #72, opened 2020-07-31
Use MS code-generator for primitives and datatypes

ewoutkramer:
The Model classes are now being generated by the Microsoft code generator, instead of the T4 templates. This means there's now a `Generated` directory inside the `Poco.Model` directory, and some primitives/datatypes have been split up into two partial classes. No other code changes.


ewoutkramer:
Note that Git got totally confused in detecting renames of the files - they were so similar (especially the primitives) that Git called some deletions and additions renames. Unfortunately this could be between two totally unrelated files.

PR #76, opened 2020-09-10
Fix NullPointer when removing the parent on importChild

marcovisserFurore:
Fixes FirelyTeam/fhir-net-api#1476


marcovisserFurore:
Approved

PR #81, opened 2020-09-21
Add string encoding funcs + calendar months to FhirPath

ewoutkramer:
The latest sets of tests published by HL7 for FhirPath include changes to the way calendar months are treated (to be more precise: it no longer uses UCUM annotations to encode calendar durations as UCUM codes).

Also, a set of experimental string-encoding functions have been added that will probably become part of the next FhirPath normative release.


marcovisserFurore (commented):
I presume you made a workaround for heximal decoding/encoding, while that function was not available for netstandard1.1 Is the function available for netstandard1.6? Because we switch now to 1.6.

PR #88, opened 2020-12-17
fix: throws error on _sort search param value that consists of just a single dash

mmsmits:
fixes #1531


marcovisserFurore (changes requested):
I missing here an unittest

mmsmits:
Unit test is in the main project

marcovisserFurore (approved):
> Unit test is in the main project

Yeah, I see that now. That unit test should have been moved to common as well, when we moved the class `SearchParams` to common. But I see that unit test has a reference to model. So it is a mixed unit test, which a pity. Anyway, I will accept it then.

PR #92, opened 2021-02-01
Fix debugger display of resource identities

mmsmits:
fixes: https://github.com/FirelyTeam/firely-net-sdk/issues/1601

Can't use extension methods in DebuggerDisplay since there is no way of knowing what using directives are in effect, and hence which extension methods are candidates


ewoutkramer:
@brianpos might want to double check whether this alternative is good enough...

brianpos:
That's super annoying, now you won't see the context that is in the Identity, such as which server it was from.
If there are safety issues with routine, probably need to resolve those, not remove it.
Or put in a DebuggerDisplay private property like in other locations to make it a little more explicit and safer.

mmsmits:
Another issue is that ResourceIdentity is still in the core project. Another PR is currently in the making to fix, after that I'll take a look at bringing this information back to the debugger.

A similar issue has occurred in StructureDefinition, will take a look at that too.

brianpos:
Ah, that's the real issue, it's not in the same assembly, or dependant...

marcovisserFurore (approved):

PR #94, opened 2021-02-08
Add utilities to parse fhir release versions

mmsmits:
fixes https://github.com/FirelyTeam/firely-net-sdk/issues/1590


src/Hl7.Fhir.Support.Poco/Utility/FhirVersionUtility.cs
```diff
@@ -0,0 +1,112 @@
...
+    public class FhirReleaseParser
+    {
+        /// <summary>
+        /// Returns a FHIR version as an enum from a version number
+        /// </summary>
+        /// <param name="version">Fhir Release version number</param>
+        /// <returns>Official FHIR Release</returns>
+        public static FhirRelease Parse(string version)
+        {
+            return version switch
```
ewoutkramer, line 17: I think this switch is too specific for "final release versions".  Take 4.0.x -> we know that this is R4, so to avoid us having to directly update the library for every single patch version.

My code looked like this:

``` csharp
 public static FhirRelease DeriveReleaseFromVersion(string fhirVersion)
        {
            if (fhirVersion.StartsWith("5.0")) return FhirRelease.R5;
            if (fhirVersion is "4.2.0" or "4.4.0" or "4.5.0") return FhirRelease.R5;
            if (fhirVersion.StartsWith("4.0")) return FhirRelease.R4;
            if (fhirVersion is "3.6.0" or "3.5a.0" or "3.5.0" or "3.4.0" or "3.3.0" or "3.2.0" or "3.1.0") return FhirRelease.R4;
            if (fhirVersion.StartsWith("3.0")) return FhirRelease.STU3;
            if (fhirVersion is "1.9.0" or "1.8.0" or "1.6.0" or "1.2.0" or "1.1.0") return FhirRelease.STU3;
            if (fhirVersion.StartsWith("1.0")) return FhirRelease.DSTU2;
            if (fhirVersion is "0.5.0" or "0.4.0") return FhirRelease.DSTU2;

            throw new NotSupportedException($"This software does not support FHIR version {fhirVersion}.");
        }
```

As you can see I switch to `StartsWith` when we know it's a final release.  Also, I noticed your code has a few more cases (especially for DSTU1), but this code also has a few more in comparison to yours.  Whatever we chose to do, we need to be as complete as possible!

src/Hl7.Fhir.Support.Poco/Utility/FhirVersionUtility.cs
```diff
@@ -0,0 +1,112 @@
...
+                _ => throw new Exception($"Unknown FHIR version {version}")
+            };
+        }
+
+        /// <summary>
+        /// Returns the version number of the latest official FHIR releases
+        /// </summary>
+        /// <param name="fhirRelease">Official FHIR release</param>
+        /// <returns>Latest version number</returns>
+        public static string FhirVersionFromRelease(FhirRelease fhirRelease)
```
ewoutkramer, line 62: I was wondering *what* we are using this for. I think I have seen it in Martijn's code, but this feels rather weird, since "the newest" version for a release is dynamic, this again would force us to push a new SDK version when there is a new FHIR patch release.  I am wondering whether Martijn uses this for creating a hard dependency or a dependency like R5 -> ">=4.5", which would give  us a bit more leeway. I'll take a look in his packaging library.

src/Hl7.Fhir.Support.Poco/Utility/FhirVersionUtility.cs
```diff
@@ -0,0 +1,112 @@
...
+                FhirRelease.DSTU1 => "0.0",
+                FhirRelease.DSTU2 => "1.0",
+                FhirRelease.STU3 => "3.0",
+                FhirRelease.R4 => "4.0",
+                FhirRelease.R5 => "5.0",
+                _ => throw new Exception($"Unknown FHIR version {fhirRelease}")
+            };
+        }
+    }
+}
```
ewoutkramer, line 112: I need one more thing:

``` csharp
  public static FhirRelease? DeriveReleaseFromCorePackageName(string packageName)
        {
            return packageName switch
            {
                "hl7.fhir.r3.core" => FhirRelease.STU3,
                "hl7.fhir.r4.core" => FhirRelease.R4,
                "hl7.fhir.r5.core" => FhirRelease.R5,
                _ => null
            };
        }
```

and maybe the reverse!

src/Hl7.Fhir.Support.Poco/Utility/FhirVersionUtility.cs
ewoutkramer: I took at look, it's unclear, since the functions that use this FhirVersionFromRelease in the packaging library don't have references in the library itself. They are probably used by Simplifier.  Well, in that case, these functions are still ok, even if they don't return the very latest fhir version given a fhir release.

ewoutkramer: How are we not going to forget that we need to keep this up-to-date?

ewoutkramer: Maybe add this update step to the process here: https://github.com/FirelyTeam/firely-net-sdk/wiki/How-to-create-a-new-release

ewoutkramer (changes requested):

ewoutkramer (changes requested):
Perfect!   Just as a reminder, can we have both a Parse() and TryParse()?  (the first throwing exceptions like now, but TryParse returning true/false?)  Martijn seems to be using that now.

src/Hl7.Fhir.Support.Poco/Utility/FhirVersionUtility.cs
```diff
@@ -44,6 +44,25 @@ public static FhirRelease Parse(string version)
...
+        /// <returns>true if the conversion succeeded; false otherwise.</returns>
+        public static bool TryParse(string version, out FhirRelease? release)
+        {
+            release = null;
+            try
+            {
+                release = Parse(version);
+                return true;
+            }
+            catch(NotSupportedException)
```
ewoutkramer, line 61: This is exactly the wrong way around ;-)  Since throwing exceptions is expensive (with all the stack unwinding etc), it's better to have the main implementation of the parser in TryParse, and then base the Parse on top of TryParse (so it throws when TryParse returns false)!

ewoutkramer (changes requested):

src/Hl7.Fhir.Support.Poco/Utility/FhirVersionUtility.cs
mmsmits: Ahhh fair enough. Will fix!

ewoutkramer (approved):

PR #98, opened 2021-02-23
Move some (partial) classes to common

marcovisserFurore:
Moved some classes from Core to Common.
- Pay extra attention to extension method `public static T Resolve<T>(this T element, Func<string, T> externalResolver = null) where T : class, ITypedElement` in class `ScopedNodeExtensions`: do we want string constants for this?
- see a list of breaking changes that this PR causes: https://github.com/FirelyTeam/firely-net-sdk/wiki/Breaking-changes-in-3.0

### DIFF
--- a/src/Hl7.Fhir.Support.Poco/ElementModel/PocoBuilderExtensions.cs
+++ b/src/Hl7.Fhir.Support.Poco/ElementModel/PocoBuilderExtensions.cs
@@ -1,0 +1,110 @@
+﻿using Hl7.Fhir.Model;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Hl7.Fhir.ElementModel
+{
+    public static class PocoBuilderExtensions
+    {
+        /// <summary>
+        /// Parses a bindeable type (code, Coding, CodeableConcept, Quantity, string, uri) into a FHIR coded datatype.
+        /// Extensions will be parsed from the 'value' of the (simple) extension.
+        /// </summary>
+        /// <param name="instance"></param>
+        /// <returns>An Element of a coded type (code, Coding or CodeableConcept) dependin on the instance type,
+        /// or null if no bindable instance data was found</returns>
+        /// <remarks>The instance type is mapped to a codable type as follows:
+        ///   'code' => code
+        ///   'Coding' => Coding
+        ///   'CodeableConcept' => CodeableConcept
+        ///   'Quantity' => Coding
+        ///   'Extension' => depends on value[x]
+        ///   'string' => code
+        ///   'uri' => code
+        /// </remarks>
+        public static Element ParseBindable(this ITypedElement instance)
+        {
+
+            return instance.InstanceType switch
+            {
+                "code" => instance.ParsePrimitive<Code>(),
+                "string" => new Code(instance.ParsePrimitive<FhirString>()?.Value),
+                "uri" => new Code(instance.ParsePrimitive<FhirUri>()?.Value),
+                "Coding" => instance.ParseCoding(),
+                "CodeableConcept" => instance.ParseCodeableConcept(),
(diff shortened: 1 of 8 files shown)


src/Hl7.Fhir.Support.Poco/ElementModel/PocoBuilderExtensions.cs
```diff
@@ -0,0 +1,110 @@
+﻿using Hl7.Fhir.Model;
```
ewoutkramer, line 1: Geen header?

src/Hl7.Fhir.Support.Poco/ElementModel/ScopedNodeExtensions.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+            // First, get the url to fetch from the focus
+            string url = null;
+
+            if (element.InstanceType == "string" && element.Value is string s) // hmm, we cannot use FhirAllTypes.String.GetLiteral()
+                url = s;
+            else if (element.InstanceType == "Reference")  // hmm, we cannot use FhirAllTypes.Reference.GetLiteral()
+                url = element.ParseResourceReference()?.Reference;
+
+            if (url == null) return default;   // nothing found to resolve
```
ewoutkramer, line 110: Yes, your question is a good one.  The agnostic bits use magic string for the most common datatypes/resources.  This will certainly improve when we are using ModelSpace, but when is that?  Should we jump and simple introduce some constants?  Where ?  I guess in Hl7.Fhir.Support.Poco?

src/Hl7.Fhir.Support.Poco/FhirPath/ElementNavFhirExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+                {
+                    return fhirValue.FhirValue;
+                }
+
+                object result = r.Value;
+
+                return result switch
+                {
+                    bool _ => new FhirBoolean((bool)result),
+                    long _ => new Integer((int)(long)result),
```
ewoutkramer, line 91: This mapping is wrong. It's int -> Integer.

ewoutkramer (changes requested):
So nice to see this stuff go....

src/Hl7.Fhir.Support.Poco/ElementModel/PocoBuilderExtensions.cs
marcovisserFurore: Added

src/Hl7.Fhir.Support.Poco/Model/FhirTypeConstants.cs
```diff
@@ -0,0 +1,51 @@
...
+ * This file is licensed under the BSD 3-Clause license
+ * available at https://github.com/FirelyTeam/firely-net-sdk/blob/master/LICENSE
+ */
+
+namespace Hl7.Fhir.Support.Poco
+{
+    /// <summary>
+    /// List of Fhir types constants. This will be removed when we introduce ModelSpace, but for now we will use this.
+    /// </summary>
+    internal class FhirTypeConstants
```
ewoutkramer, line 14: `internal` - this means other assemblies have to re-do this work. On the other hand, if you make this public, "why didn't you add a constant for X"  (because *we* didn't need it ;-))......

ewoutkramer (approved):

src/Hl7.Fhir.Support.Poco/Model/FhirTypeConstants.cs
marcovisserFurore: Internal, because when we introduce ModelSpace we can safely remove this class without breaking the code.

PR #103, opened 2021-03-22
Support relative URLs on location headers when parsing HttpResponseMessage to EntryResponse

mmsmits:
fixes https://github.com/FirelyTeam/firely-net-sdk/issues/1659

### DIFF
--- a/src/Hl7.Fhir.Support.Poco/Rest/HttpToEntryExtensions.cs
+++ b/src/Hl7.Fhir.Support.Poco/Rest/HttpToEntryExtensions.cs
@@ -30,7 +30,7 @@ namespace Hl7.Fhir.Rest
                 Status = ((int)response.StatusCode).ToString(),
                 ResponseUri = response.RequestMessage.RequestUri,//this is actually the requestUri, can't find the responseUri
                 Body = body,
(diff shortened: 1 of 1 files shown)


src/Hl7.Fhir.Support.Poco/Rest/HttpToEntryExtensions.cs
```diff
@@ -30,7 +30,7 @@ internal static EntryResponse ToEntryResponse(this HttpResponseMessage response,
                 Status = ((int)response.StatusCode).ToString(),
                 ResponseUri = response.RequestMessage.RequestUri,//this is actually the requestUri, can't find the responseUri
                 Body = body,
-                Location = response.Headers.Location?.AbsoluteUri ?? response.Content.Headers.ContentLocation?.AbsoluteUri,
+                Location = response.Headers.Location?.ToString() ?? response.Content.Headers.ContentLocation?.ToString(),
```
marcovisserFurore, line 33: Better to use OriginalString. See also some background at https://docs.microsoft.com/en-us/dotnet/api/system.uri.originalstring?view=net-5.0#remarks

marcovisserFurore (changes requested):
See comment

marcovisserFurore (approved):

PR #108, opened 2021-03-31
Adding setting Serialize settings IncludeMandatoryToElements

marcovisserFurore:
resolves FirelyTeam/firely-net-sdk#1653

### DIFF
--- a/src/Hl7.Fhir.Support.Poco/Serialization/SerializerSettings.cs
+++ b/src/Hl7.Fhir.Support.Poco/Serialization/SerializerSettings.cs
@@ -30,9 +30,9 @@ namespace Hl7.Fhir.Serialization
         public bool TrimWhiteSpacesInXml { get; set; } = true;
 
         /// <summary>
-        /// Include mandatory elements to the given set elements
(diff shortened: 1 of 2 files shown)


src/Hl7.Fhir.Support.Poco/Serialization/SerializerSettings.cs
```diff
@@ -29,6 +29,11 @@ public class SerializerSettings
         /// </summary>
         public bool TrimWhiteSpacesInXml { get; set; } = true;
 
+        /// <summary>
+        /// Include mandatory elements to the given set elements
+        /// </summary>
+        public bool IncludeMandatoryToElements { get; set; } = false;
```
ewoutkramer, line 35: Nothing technical, but I did not "get" the name when I read it.  I'll try a few alternatives:

* IncludeMandatoryInElementsSummary
* ElementsShouldIncludeMandatory
* IncludeMandatoryWithElements

The comment:  "Include mandatory elements when serializing a subset of chosen elements (_elements). See <see cref="Method with elements overload" />.

ewoutkramer (changes requested):

ewoutkramer (approved):

PR #112, opened 2021-04-30
Async support Json/Xml Parsers + Serializers

kennethmyhra:
Async support for Json Parsers related and Json/Xml Serializers

Non-async methods marked as obsolete, documentation refers to the corresponding Async method

Corresponding PR https://github.com/FirelyTeam/firely-net-sdk/pull/1700

### DIFF
--- a/src/Hl7.Fhir.Serialization/FhirXmlBuilderExtensions.cs
+++ b/src/Hl7.Fhir.Serialization/FhirXmlBuilderExtensions.cs
@@ -9,6 +9,8 @@
 
 using Hl7.Fhir.ElementModel;
 using Hl7.Fhir.Utility;
+using System;
+using System.Threading.Tasks;
 using System.Xml;
 using System.Xml.Linq;
 
@@ -16,6 +18,8 @@ namespace Hl7.Fhir.Serialization
 {
     public static class FhirXmlBuilderExtensions
     {
+        /// <inheritdoc cref="writeToAsync(XDocument, XmlWriter)" />
+        [Obsolete("Use writeToAsync(XDocument, XmlWriter) instead.")]
         private static void writeTo(this XDocument doc, XmlWriter destination)
         {
             if (doc.Root != null)
@@ -24,11 +28,29 @@ namespace Hl7.Fhir.Serialization
             destination.Flush();
         }
 
+        private static async Task writeToAsync(this XDocument doc, XmlWriter destination)
+        {
+            if (doc.Root != null)
+                doc.WriteTo(destination);
+
+            await destination.FlushAsync().ConfigureAwait(false);
+        }
+
+        /// <inheritdoc cref="WriteToAsync(ISourceNode, XmlWriter, FhirXmlSerializationSettings)" />
+        [Obsolete("Use WriteToAsync(ISourceNode, XmlWriter, FhirXmlSerializationSettings) instead.")]
         public static void WriteTo(this ISourceNode source, XmlWriter destination, FhirXmlSerializationSettings settings = null) =>
             new FhirXmlBuilder(settings).Build(source).writeTo(destination);
 
+        public static async Task WriteToAsync(this ISourceNode source, XmlWriter destination, FhirXmlSerializationSettings settings = null) =>
+            await new FhirXmlBuilder(settings).Build(source).writeToAsync(destination).ConfigureAwait(false);
+
+        /// <inheritdoc cref="WriteToAsync(ITypedElement, XmlWriter, FhirXmlSerializationSettings)" />
+        [Obsolete("Use WriteToAsync(ITypedElement, XmlWriter, FhirXmlSerializationSettings) instead.")]
         public static void WriteTo(this ITypedElement source, XmlWriter destination, FhirXmlSerializationSettings settings = null) =>
             new FhirXmlBuilder(settings).Build(source).writeTo(destination);
+
+        public static async Task WriteToAsync(this ITypedElement source, XmlWriter destination, FhirXmlSerializationSettings settings = null) =>
+            await new FhirXmlBuilder(settings).Build(source).writeToAsync(destination).ConfigureAwait(false);
 
         public static XDocument ToXDocument(this ISourceNode source, FhirXmlSerializationSettings settings = null) =>
             new FhirXmlBuilder(settings).Build(source);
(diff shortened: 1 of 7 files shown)


src/Hl7.Fhir.Serialization/FhirXmlBuilderExtensions.cs
```diff
@@ -24,25 +28,58 @@ private static void writeTo(this XDocument doc, XmlWriter destination)
             destination.Flush();
         }
 
+        private static async Task writeToAsync(this XDocument doc, XmlWriter destination)
+        {
+            if (doc.Root != null)
+                doc.WriteTo(destination);
```
marcovisserFurore, line 34:
```suggestion
                await doc.WriteToAsync(destination).ConfigureAwait(false);
```

marcovisserFurore (changes requested):
Nice work, but I am not so happy that we make all the 'sync' methods obsolete. That means that all our end users should upgrade their code to async, which is not so easy and can take a lot of time.

My suggestion would be to have sync and async methods. The end user can then still decide which one to take. Furthermore, I would suggest to have a bodyless sync method which only contains a call to the corresponding async method.  

For example:
```c#
private static void writeTo(this JObject root, JsonWriter destination, string rootName = null)
    => TaskHelper.Await(() => writeToAsync(root,  destination,  rootName ));
		

private static async Task writeToAsync(this JObject root, JsonWriter destination, string rootName = null)
{
	await root.WriteToAsync(destination);
	await destination.FlushAsync();
}	
```

kennethmyhra:
I agree, I will remove the Obsolete attributes. Also will add async methods for the XML parser to align it with the Json parser.

src/Hl7.Fhir.Serialization/FhirXmlBuilderExtensions.cs
kennethmyhra: WriteToAsync is only available in netstandard2.0. We are supporting both netstandard1.3 and .NET Framework 4.5

brianpos:
Have you done any performance comparisons between the sync/async code here?
(I found the async version (in my new parser/serializer) was significantly slower)
And there is also a bug in the MS XML async readers that you'll want to be careful of not tripping over on - I have a workaround for it in my code if you get to it. (one of their async functions calls a sync function internally)

kennethmyhra:
I have not done any performance tests, but we will keep the sync-methods.

Do share the workaround for the bug so I can have a look at it and see if it is relevant for this case as well. Thread locking concerns when mixing async and sync should only be an issue with .NET FW on desktop and web applications as far as I have understood.

src/Hl7.Fhir.Serialization/FhirXmlBuilderExtensions.cs
kennethmyhra: Available in netstandard2.1 actually

marcovisserFurore: Ah ok, I was not aware of that.

brianpos:
https://github.com/brianpos/fhir-net-web-api/blob/e324bb3a6f9a2cc98b63be6c187deb4d868e097d/src/Hl7.Fhir.Custom.Serializers/FhirCustomXmlReader.core.cs#L318

kennethmyhra:
Does not look like we call `ReadOuterXmlAsync` from the library. We are passing the `XmlReader` to `XDocument.Load` and then we pass `XDocument.Root` to the constructor of `FhirXmlNode`

`XDocument.Load` should load the whole XML document into memory. If we experience any problems in the future it will be here. Unfortunately `XDocument.LoadAsync` is only available in netstandard2.1 + netcore2.0 and later.

src/Hl7.Fhir.Serialization/Utility/SerializationUtil.cs
```diff
@@ -167,6 +228,27 @@ public static string WriteXmlToString(Action<XmlWriter> serializer, bool pretty
...
+            // [WMR 20160421] Explicit disposal
+            using (XmlWriter xw = XmlWriter.Create(sb, settings))
+            {
+                await serializer(xw).ConfigureAwait(false);
+                await xw.FlushAsync().ConfigureAwait(false);
+                return appendNewLine ? sb.AppendLine().ToString() : sb.ToString();
+            }
+        }
+
         public static XDocument WriteXmlToDocument(Action<XmlWriter> serializer)
```
marcovisserFurore, line 252: Maybe also create a async version of this method (`WriteXmlToDocument`)?

marcovisserFurore (changes requested):
Some minor things

kennethmyhra:
Fixed

src/Hl7.Fhir.Serialization/Utility/SerializationUtil.cs
```diff
@@ -261,6 +262,19 @@ public static XDocument WriteXmlToDocument(Action<XmlWriter> serializer)
...
         }
+        
+        public static async Task<XDocument> WriteXmlToDocumentAsync(Func<XmlWriter, Task> serializer)
+        {
+            var doc = new XDocument();
+
+            using (XmlWriter xw = doc.CreateWriter())
+            {
+                await serializer(xw).ConfigureAwait(false);
+                await xw.FlushAsync();
```
marcovisserFurore, line 273:
```suggestion
                await xw.FlushAsync().ConfigureAwait(false);
```

marcovisserFurore (changes requested):
Missing .ConfigureAwait(false)

marcovisserFurore (approved):

kennethmyhra:
Performance tests sync vs async methods using the FullRoundtripOfAllExamples*. As expected there is a slight overhead for the async methods.

Asynchronous programming isn't about pure performance like these methods tests, among other things it is about scalability

FullRoundtripOfAllExamplesJsonNavPocoProvider:
```
Roundtripping json->xml->json
(...)
Conversion of 1664 files took 16 seconds
-
Re-converting files in (...) back to original format
(...)
Conversion of 1664 files took 12 seconds
```

FullRoundtripOfAllExamplesJsonNavPocoProviderAsync:
```
Roundtripping json->xml->json
Converting files in (...)
(...)
Conversion of 1664 files took 16 seconds
-
Re-converting files in (...) back to original format
(...)
Conversion of 1664 files took 12 seconds
```

FullRoundtripOfAllExamplesJsonNavSdProvider:
```
Roundtripping json->xml->json
Converting files in (...)
(...)
Conversion of 1664 files took 24 seconds
-
Re-converting files in (...) back to original format
(...)
Conversion of 1664 files took 19 seconds
```

FullRoundtripOfAllExamplesJsonNavSdProviderAsync:
```
Roundtripping json->xml->json
Converting files in (...)
(...)
Conversion of 1664 files took 27 seconds
-
Re-converting files in (...) back to original format
(...)
Conversion of 1664 files took 20 seconds
```

FullRoundtripOfAllExamplesJsonPoco:
```
Roundtripping json->xml->json
Converting files in (...)
(...)
Conversion of 1664 files took 18 seconds
-
Re-converting files in (...) back to original format
(...)
Conversion of 1664 files took 28 seconds
```

FullRoundtripOfAllExamplesJsonPocoAsync:
```
Roundtripping json->xml->json
Converting files in (...)
(...)
Conversion of 1664 files took 19 seconds
-
Re-converting files in (...) back to original format
(...)
Conversion of 1664 files took 30 seconds
```

FullRoundtripOfAllExamplesXmlNavPocoProvider:
```
Roundtripping xml->json->xml
Converting files in (...)
(...)
Conversion of 1664 files took 15 seconds
-
Re-converting files in (...) back to original format
(...)
Conversion of 1664 files took 13 seconds
```

FullRoundtripOfAllExamplesXmlNavPocoProviderAsync:
```
Roundtripping xml->json->xml
Converting files in (...)
(...)
Conversion of 1664 files took 16 seconds
-
Re-converting files in (...) back to original format
(...)
Conversion of 1664 files took 14 seconds
```

FullRoundtripOfAllExamplesXmlNavSdProvider:
```
Roundtripping xml->json->xml
Converting files in (...)
(...)
Conversion of 1664 files took 24 seconds
-
Re-converting files in (...) back to original format
(...)
Conversion of 1664 files took 18 seconds
```

FullRoundtripOfAllExamplesXmlNavSdProviderAsync:
```
Roundtripping xml->json->xml
Converting files in (...)
(...)
Conversion of 1664 files took 24 seconds
-
Re-converting files in (...) back to original format
(...)
Conversion of 1664 files took 18 seconds
```

FullRoundtripOfAllExamplesXmlPoco:
```
Roundtripping xml->json->xml
Converting files in (...)
(...)
Conversion of 1664 files took 34 seconds
-
Re-converting files in (...) back to original format
(...)
Conversion of 1664 files took 16 seconds
```

FullRoundtripOfAllExamplesXmlPocoAsync:
```
Roundtripping xml->json->xml
Converting files in (...)
(...)
Conversion of 1664 files took 33 seconds
-
Re-converting files in (...) back to original format
(...)
Conversion of 1664 files took 16 seconds
```

src/Hl7.Fhir.Support.Tests/JsonAssert.cs
```diff
@@ -19,8 +19,10 @@ public class JsonAssert
     {
         public static void AreSame(string filename, string expected, string actual, List<string> errors)
         {
+#pragma warning disable CS0618 // Type or member is obsolete
```
ewoutkramer, line 22: Minor: I don't think these methods are marked `Obsolete` anymore, so these pragma's can be removed?

ewoutkramer (approved):

ewoutkramer (approved):

PR #113, opened 2021-05-03
Move another set of classes to Common

ewoutkramer:
Another set of classes that can now be moved to Common


marcovisserFurore (approved):
Checked with PR https://github.com/FirelyTeam/firely-net-sdk/pull/1701

PR #118, opened 2021-05-17
Keep existing timezone by default in FhirDateTime.ToDateTimeOffset - Common

mmsmits:
fixes: https://github.com/FirelyTeam/firely-net-sdk/issues/1498


marcovisserFurore (changes requested):
Today discussed in our standup:

The method `ToDateTimeOffset()` without a `zone` was explicit set to Obsolete. 

It would be better (also to support the user of the issue) to create a new function:

`public bool TryToDateTimeOffset(out DateTimeOffset dto)`

which returns `true` and a dto when we can determine a zone, `false` otherwise.

In the case of true, you can determine the dto by forwarding it to  `ToDateTimeOffset(Timespan zone)`.

src/Hl7.Fhir.Support.Poco/Model/FhirDateTime.cs
```diff
@@ -107,6 +107,77 @@ public DateTimeOffset ToDateTimeOffset(TimeSpan zone)
...
+            }
+            else
+            {
+                return false;
+            }
+        }
+
+        private bool containsTimeZone()
+        {
+            //YYYY-MM-DDThh:mm:ss.sss+zz:zz //
```
marcovisserFurore, line 168: The regex of a `FhirDateTime` is:
```
([0-9]([0-9]([0-9][1-9]|[1-9]0)|[1-9]00)|[1-9]000)(-(0[1-9]|1[0-2])(-(0[1-9]|[1-2][0-9]|3[0-1])(T([01][0-9]|2[0-3]):[0-5][0-9]:([0-5][0-9]|60)(\.[0-9]+)?(Z|(\+|-)((0[0-9]|1[0-3]):[0-5][0-9]|14:00)))?)?)?
```

So the fraction after the second, can be very precise. For example this is also a correct FhirDateTime:

```
2021-03-18T12:22:35.12345+04:00
```

So the `+` or `-` of the Timezone is not always on position 23 in the string.

Would it not be an idea to parse the string with the regex? Then you can extract the timezone from that.

marcovisserFurore: You can get some inspiration from this code: https://github.com/FirelyTeam/firely-net-common/blob/115fce4c1fab236fd80685aa1030ba82be925edd/src/Hl7.Fhir.ElementModel/Types/DateTime.cs#L93

marcovisserFurore (changes requested):
Fraction can be longer than 3 positions

marcovisserFurore (approved):

PR #121, opened 2021-05-21
Refreshed some metadata classes in preparation for using assemblies as metadata instead of ISDSP

ewoutkramer:


src/Hl7.Fhir.Support.Poco/ElementModel/TypedSerialization.cs
```diff
@@ -10,53 +10,28 @@
...
-using System.Collections.Concurrent;
 using System.Reflection;
 
 namespace Hl7.Fhir.ElementModel
 {
     public static class TypedSerialization
     {
-        private static readonly ConcurrentDictionary<string, ModelInspector> _inspectedAssemblies = new();
-
-        public static ModelInspector GetInspectorForAssembly(Assembly a)
```
marcovisserFurore, line 22: Do we have here a breaking change? Or was this method not release yet? I am not sure anymore..

marcovisserFurore (changes requested):
See my question

src/Hl7.Fhir.Support.Poco/ElementModel/TypedSerialization.cs
ewoutkramer: We haven't released this code yet. It's just 5 days old!

marcovisserFurore: Then we can pull it. Thanks

marcovisserFurore (approved):

PR #134, opened 2021-07-06
Added ProfileRelease configuration

GinoCanessa:
Added a new build configuration to allow building release without signing.
* added the new configuration to firely-net-common.props
* added the new configuration to the solution file
* configuration defines variable PROFILE_RELEASE
* added PROFILE_RELEASE to `InternalVisibleTo` compiler directives to allow weak naming


GinoCanessa:
Moving PR to different target branch, per request.

PR #135, opened 2021-07-07
FHIR Specific variables %rootResource and %resource are set correctly now

marcovisserFurore:
FHIR Specific variables `%rootResource` and `%resource` are now set correctly. See also the assumptions in https://github.com/FirelyTeam/firely-net-sdk/issues/1742#issuecomment-875598124.

Resolves FirelyTeam/firely-net-sdk#1742

### DIFF
--- a/src/Hl7.Fhir.Support.Poco/FhirPath/FhirEvaluationContext.cs
+++ b/src/Hl7.Fhir.Support.Poco/FhirPath/FhirEvaluationContext.cs
@@ -32,6 +32,14 @@ namespace Hl7.Fhir.FhirPath
         {
         }
 
+        public FhirEvaluationContext(ScopedNode resource)
+            : base(toNearestParentResource(resource))
+        {
+            RootResource = Resource is ScopedNode node ? node.ResourceContext : resource;
+        }
+
+        private static ScopedNode toNearestParentResource(ScopedNode node) => node.AtResource ? node : node.ParentResource;
+
         private Func<string, ITypedElement> _elementResolver;
 
         public Func<string, ITypedElement> ElementResolver
--- a/src/Hl7.Fhir.Support.Poco/FhirPath/FhirEvaluationContext.cs
+++ b/src/Hl7.Fhir.Support.Poco/FhirPath/FhirEvaluationContext.cs
@@ -33,12 +33,12 @@ namespace Hl7.Fhir.FhirPath
         }
 
         public FhirEvaluationContext(ScopedNode resource)
-            : base(toNearestParentResource(resource))
+            : this(toNearestParentResource(resource))
         {
             RootResource = Resource is ScopedNode node ? node.ResourceContext : resource;
         }
 
-        private static ScopedNode toNearestParentResource(ScopedNode node) => node.AtResource ? node : node.ParentResource;
+        private static ITypedElement toNearestParentResource(ScopedNode node) => node.AtResource ? node : node.ResourceContext;
 
         private Func<string, ITypedElement> _elementResolver;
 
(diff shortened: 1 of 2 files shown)


ewoutkramer:
Above the test for this code I wrote:

> %resource and %rootResource don't really work well unless run from within the validator. " +
            "Need to fix this by letting the FP evaluator pick up these context variables from the ScopedNode focus." +
            "But that's not yet the case."

The logic to determine these variables in the current PR is equivalent to the exact same code in the validator (https://github.com/FirelyTeam/firely-net-sdk/blob/develop-stu3/src/Hl7.Fhir.Specification/Validation/FpConstraintValidationExtensions.cs#L58) that we should now remove. *But* it is not clear from the public interface of the constructor for the EvaluationContext that it will "reinterpret" the parameters passed to it.  So, in the case of the validator, the same determination process is now done twice - quite unexpected.  Maybe it is better to keep the current constructor as it is, and add a constructor for the EvaluationContext that takes an explicit ScopedNode - and then derives both variables from it just like the code does in the validator / current PR.
 
> The FP variables %resource and %rootResource are currently just static variables passed in via the EvaluationContext. However, the nature of these variables is that these variables are dynamic and will change as you walk into, say, a bundle, into an entry and then into a contained resource. This currently does not happen.

I think we need to agree whether or not this PR solves this.  When I wrote the above I was convinced that %resource and %rootResource would *vary* during evaluation, so if you move into a resource in FhirPath, %resource would change. Clearly, you (and myself when I wrote the old code) did not think so.  So, maybe indeed %resource and %rootResource are determined based on the context in which the FP statement is invoked (the focus at that time), and does not change while evaluating the FP statement.  Let's agree that the latter is indeed the case.  If they would be more dynamic, they probably should have looked like a function (e.g. `parentResource()`) instead of an external variable (`%xxxxxxx`).

ewoutkramer (changes requested):
I'd like to add a new constructor instead of changing the old one...

marcovisserFurore:
> I'd like to add a new constructor instead of changing the old one...

A new constructor on `EvaluationContext` or on `FhirEvaluationContext` or on both?

ewoutkramer:
> > I'd like to add a new constructor instead of changing the old one...
> 
> A new constructor on `EvaluationContext` or on `FhirEvaluationContext` or on both?

Since %resource and %rootResource are FHIR specific, I'd say just on the FhirEvaluationContext.

ewoutkramer (dismissed):

src/Hl7.Fhir.Support.Poco/FhirPath/FhirEvaluationContext.cs
```diff
@@ -33,12 +33,12 @@ public FhirEvaluationContext(ITypedElement resource, ITypedElement rootResource)
...
 
         public FhirEvaluationContext(ScopedNode resource)
-            : base(toNearestParentResource(resource))
+            : this(toNearestParentResource(resource))
         {
             RootResource = Resource is ScopedNode node ? node.ResourceContext : resource;
         }
 
-        private static ScopedNode toNearestParentResource(ScopedNode node) => node.AtResource ? node : node.ParentResource;
+        private static ITypedElement toNearestParentResource(ScopedNode node) => node.AtResource ? node : node.ResourceContext;
```
ewoutkramer: The name of this function is now pretty misleading ;-)

ewoutkramer (changes requested):
Please, just pull when you're done ;-)

src/Hl7.Fhir.Support.Poco/FhirPath/FhirEvaluationContext.cs
marcovisserFurore: I think you looked at the a previous commit. It is called now `toNearestResource`. See https://github.com/FirelyTeam/firely-net-common/blob/cf4db7cbd5475647abd1bde4725e8fef981ebdf1/src/Hl7.Fhir.Support.Poco/FhirPath/FhirEvaluationContext.cs#L45

PR #136, opened 2021-07-08
Move from netstandard2.1 to NET 5.0

mmsmits:


marcovisserFurore (approved):
Although I am in favor of solving the warnings (if possible) rather than suppressing them, I approve this PR. On the board there is now an [action ](https://github.com/FirelyTeam/firely-net-sdk/projects/31#card-64695934)to try to resolved the nullable warnings.

PR #139, opened 2021-07-13
Solved the nullable warnings

marcovisserFurore:
With the introduction of `net5.0` we suppressed some nullable warnings. These have been solved now

### DIFF
--- a/src/Hl7.Fhir.Support.Poco/Introspection/ModelInspector.cs
+++ b/src/Hl7.Fhir.Support.Poco/Introspection/ModelInspector.cs
@@ -6,8 +6,7 @@
  * available at https://raw.githubusercontent.com/FirelyTeam/firely-net-sdk/master/LICENSE
  */
 
-#nullable enable annotations
-#nullable disable warnings
+#nullable enable
 
 using Hl7.Fhir.Model;
 using Hl7.Fhir.Specification;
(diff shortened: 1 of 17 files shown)


src/Hl7.Fhir.Support.Poco/Introspection/ModelInspector.cs
```diff
@@ -42,7 +41,7 @@ public class ModelInspector : IStructureDefinitionSummaryProvider
         /// metadata for the most recent release for those common classes.</remarks>
         public static ModelInspector ForAssembly(Assembly a)
         {
-            return _inspectedAssemblies.GetOrAdd(a.FullName, _ => configureInspector(a));
+            return _inspectedAssemblies.GetOrAdd(a.GetName().FullName, _ => configureInspector(a));
```
ewoutkramer, line 44: Sure? Is that the same?

src/Hl7.Fhir.Support.Poco/Introspection/PropertyMapping.cs
```diff
@@ -204,7 +203,7 @@ private static bool isPrimitiveValueElement(FhirElementAttribute valueElementAtt
 
         }
 
-        private static string buildQualifiedPropName(PropertyInfo p) => p.DeclaringType.Name + "." + p.Name;
+        private static string buildQualifiedPropName(PropertyInfo p) => p.DeclaringType?.Name + "." + p.Name;
```
ewoutkramer, line 206: Actually, if this is null, we have a problem and should throw.

ewoutkramer (changes requested):

src/Hl7.Fhir.Support.Poco/Introspection/ModelInspector.cs
marcovisserFurore: I reversed the the change, and throw an `ArgumentNullException` when `a.FullName` is null.

src/Hl7.Fhir.Support.Poco/Introspection/PropertyMapping.cs
marcovisserFurore: I throw now an `ArgumentNullException`.

ewoutkramer (approved):

PR #141, opened 2021-07-19
Adding IReadOnlyDictionary support

ewoutkramer:
Includes changes to generated classes and some manual implementations.


src/Hl7.Fhir.Support.Poco/Model/PocoDictionary.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        public static string ComposeChoiceElementName(string baseName, Base elementValue) =>
+            baseName + char.ToUpperInvariant(elementValue.TypeName[0]) + elementValue.TypeName.Substring(1);
+
+        public static string ComposeChoiceElementName(string baseName, IEnumerable<Base> elementValues)
+        {
+            var firstValue = elementValues.FirstOrDefault();
+
+            return firstValue is not null
+                ? baseName + char.ToUpperInvariant(firstValue.TypeName[0]) + firstValue.TypeName.Substring(1)
```
marcovisserFurore, line 25: perhaps call the `ComposeChoiceElementName(string, Base)` here? Now we have slightly duplicate code.

src/Hl7.Fhir.Support.Poco/Model/PocoDictionary.cs
```diff
@@ -0,0 +1,32 @@
...
+                : throw new InvalidOperationException($"Cannot serialize an empty array at element '{baseName}'.");
+        }
+
+        public static bool HasCorrectSuffix(string elementName, string expectedSuffix, int elementLength) =>
```
marcovisserFurore, line 29: I presume it is faster to calculate the `elementLength` during generating code, then doing it here, right?

marcovisserFurore (changes requested):
See my comment about duplicate code. I leave it up to you to change it or not.

src/Hl7.Fhir.Support.Poco/Model/PocoDictionary.cs
ewoutkramer: Yes, indeed. In fact, I'll move these methods to a public utility class so they can be used by others implementing IReadOnlyDictionary as well.

src/Hl7.Fhir.Support.Poco/Model/PocoDictionary.cs
ewoutkramer: Yes, but it turned out to be minimal after I did some additional profiling, so I have changed to signature to not take a constant int but the element name itself.

marcovisserFurore: That makes it more readable I think. When I first saw it, I was thinking what this integer was doing there.

marcovisserFurore (approved):

ewoutkramer:
Closes https://github.com/FirelyTeam/firely-net-sdk/issues/554

PR #151, opened 2021-09-01
Added netstandard2.0 as one of the TargetFrameworks

marcovisserFurore:
[Recommended](https://docs.microsoft.com/en-us/dotnet/standard/net-standard#which-net-standard-version-to-target) by Microsoft to add `netstandard2.0` as a TargetFramework as well. Also when you already target `netstandard1.x`

### DIFF
--- a/src/Hl7.Fhir.Support/Hl7.Fhir.Support.csproj
+++ b/src/Hl7.Fhir.Support/Hl7.Fhir.Support.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 	<PropertyGroup>
-		<TargetFrameworks>net5.0;net45;netstandard1.6</TargetFrameworks>
+		<TargetFrameworks>net5.0;net45;netstandard1.6;netstandard2.0</TargetFrameworks>
 	</PropertyGroup>
 
 	<Import Project="..\firely-net-common.props" />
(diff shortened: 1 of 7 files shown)


src/Hl7.Fhir.Support/Hl7.Fhir.Support.csproj
```diff
@@ -15,7 +15,7 @@
 		<AssemblyName>Hl7.Fhir.Support</AssemblyName>
 	</PropertyGroup>
 
-	<ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.6' or '$(TargetFramework)' == 'net5.0'">
```
mmsmits, line 18: I don't think this library is required by net 45, so probably better to change it to:
`'$(TargetFramework)' == 'netstandard1.6' or $(TargetFramework)' == 'netstandard2.0' or '$(TargetFramework)' == 'net5.0'`

mmsmits (changes requested):
One small discussionpoint

src/Hl7.Fhir.Support/Hl7.Fhir.Support.csproj
marcovisserFurore: Or `'$(TargetFramework)' != 'net45`

mmsmits (approved):

PR #153, opened 2021-09-10
FhirPath function htmlChecks() checks now also for non-whitespace content

marcovisserFurore:
FhirPath function `htmlChecks` checks now also for non-whitespace content (text or an image).


mmsmits (approved):
Nice, DataTestMethod

PR #155, opened 2021-09-20
Add NpmPackageResolver

mmsmits:


marcovisserFurore (changes requested):
Add the project also to the solution, then the build server can process it as well.

Further reviewing will follow

src/Firely.Fhir.Packages/Utils/Utils.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            foreach (var item in other)
+            {
+                dictionary.Add(item.Key, item.Value);
+            }
+        }
+
+        public static bool IsValidUrl(string source)
+        {
+#if !NETSTANDARD1_6
```
marcovisserFurore, line 32: For NETSTANDARD1_6 I would define 2 string constants that contain respectively  `https` and `http` and then use them here, so the function `IsValidUrl` still can be used

src/Firely.Fhir.Packages/Tar/Tar.cs
```diff
@@ -0,0 +1,235 @@
...
+
+            using var file = File.Create(packagefile);
+            using var gzip = new GZipOutputStream(file);
+            using TarOutputStream tar = new TarOutputStream(gzip);
+
+            Write(tar, entries);
+
+            return packagefile;
+#else
+            return "";
```
marcovisserFurore, line 28: To be uniform with the other methods, I would throw here also a `NotImplementedException`

marcovisserFurore (changes requested):
I manually compared the changes with original code, so I could see what was changed.

The `azure-pipelines.yml` should also be changed, because now the `Firely.Fhir.Packages.Tests` do not run now.  Add `**/publish/Firely.Fhir.Packages.Tests.dll` to the projects at the end of the file.

Furthermore, because this is new code to our SDK solution, we should enable nullable as well. But that could also be a task for polishing this project in the near future.

marcovisserFurore (approved):
The project `Firely.Fhir.Packages` has been added to the SDK and included in the Azure Pipelines. Also it can now be compiled against `netstandard1.6`.

The following steps that we still need to make are:
- reduce the number of public methods/classes
- polish the code to make it publishable.

src/Hl7.Fhir.Support.Poco/Specification/Source/NpmPackageResolver.cs
```diff
@@ -0,0 +1,199 @@
...
+        private ModelInspector _provider;
+
+        /// <summary>Create a new <see cref="NpmPackageResolver"/> instance to read FHIR artifacts from one or multiple FHIR packages of a specific FHIR version
+        /// found in the paths passed to this function.</summary>
+        /// <returns>A new <see cref="NpmPackageResolver"/> instance.</returns>
+        /// <param name="provider">A <see cref="ModelInspector"/> used to parse the filecontents to FHIR resources, this is typically a <see cref="ModelInspector"/> containing the definitions of a specific FHIR version. </param>
+        /// <param name="filePaths">A path to the FHIR package files.</param>
+        public NpmPackageResolver(ModelInspector provider, params string[] filePaths)
+        {
+            _context = TaskHelper.Await(() => createPackageContextAsync(filePaths));
```
marcovisserFurore, line 31: Would it not be better to make this a Lazy initialization? Constructors don't have to do the heavy lifting themselves.

src/Hl7.Fhir.Support.Poco/Specification/Source/NpmPackageResolver.cs
```diff
@@ -0,0 +1,199 @@
...
+        }
+
+
+        private static PackageContext createContext(string folder, bool localCache = false)
+        {
+            string? cache_folder = localCache ? folder : null;
+            var cache = new DiskPackageCache(cache_folder);
+            var project = new FolderProject(folder);
+#pragma warning disable CS8625 // Cannot convert null literal to non-nullable reference type.
+            return new PackageContext(cache, project, null);
```
marcovisserFurore, line 140: So the third argument of the PackageContext constructor (`IPackageServer server`) should be made nullable

src/Hl7.Fhir.Support.Poco/Specification/Source/NpmPackageResolver.cs
```diff
@@ -0,0 +1,199 @@
...
+        {
+            var content = await _context.GetCodeSystemByValueSet(valueSetUri);
+            return content is null ? null : toResource(content);
+        }
+
+        ///<inheritdoc/>
+        public async Task<IEnumerable<Resource>?> FindConceptMaps(string? sourceUri = null, string? targetUri = null)
+        {
+            var content = await _context.GetConceptMapsBySourceAndTarget(sourceUri, targetUri);
+            if (content is null)
```
marcovisserFurore, line 181: You can write it like this to avoid the nullability warning:

```c#
return content is null
                ? null
                : (from item in content
                       let resource = toResource(item)
                       where resource is not null
                       select resource);
```

marcovisserFurore (changes requested):
Perhaps we can find a way to to test this class also in Common. Ewout did a few test with POCO's in the serializer/deserializer

src/Hl7.Fhir.Support.Poco/Specification/Source/NpmPackageResolver.cs
mmsmits: Nice!

src/Hl7.Fhir.Support.Poco/Specification/Source/NpmPackageResolver.cs
mmsmits: Yes, in the package library. I'll fix that.

marcovisserFurore (approved):
Instead of using `_context.Value.<etc>` you can also create a private property, which does that for you:
```c#
  PackageContext Context => _context.Value;
```

PR #157, opened 2021-09-23
TypedElementOnSourceNode: Caching of Children and Value to improve performance

marcovisserFurore:
|                     Method |     Mean |    Error |   StdDev |      Gen 0 |     Gen 1 | Gen 2 | Allocated |
|--------------------------- |---------:|---------:|---------:|-----------:|----------:|------:|----------:|
| ToTypedElementOnSourceNode (before caching)| 885.8 ms | 26.63 ms | 77.26 ms | 23000.0000 | 3000.0000 |     - |    143 MB |
| ToTypedElementOnSourceNode (after caching)| 724.1 ms | 14.34 ms | 20.56 ms | 19000.0000 | 9000.0000 |     - |    117 MB |

### DIFF
--- a/src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
+++ b/src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
@@ -20,6 +20,9 @@ namespace Hl7.Fhir.ElementModel
         private const string XHTML_INSTANCETYPE = "xhtml";
         private const string XHTML_DIV_TAG_NAME = "div";
 
+        private readonly Lazy<object> _value;
+        private readonly Lazy<IEnumerable<ITypedElement>> _children;
+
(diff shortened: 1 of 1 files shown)


src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
```diff
@@ -74,6 +77,7 @@ private TypedElementOnSourceNode(TypedElementOnSourceNode parent, ISourceNode so
             Definition = definition;
             InstanceType = instanceType;
             _settings = parent._settings;
+            _value = new Lazy<object>(valueFactory);
```
ewoutkramer, line 80: This will cause every source node to have a Lazy<T>() allocated, even if the .Value property is never called. Can we use https://docs.microsoft.com/en-us/dotnet/api/system.threading.lazyinitializer.ensureinitialized?view=net-5.0 instead?

ewoutkramer (changes requested):

PR #159, opened 2021-10-11
Aggregate function implementation

brianpos:
Implementation of the $aggregate fhirpath function (the TestPublishedTests has unit tests for this - in the FhirPathFromSpec category)


marcovisserFurore:
The HL7 provided FhirPath tests are in the core project and only available in the R4 branch.

PR #163, opened 2021-10-21
Packaging: Move index file to root

mmsmits:
Index.json now also indexes files in sibling folders.


mmsmits:
merged after Martijns approval

PR #168, opened 2021-11-16
Five W's support

itye-msft:
Following the discussion here: [https://github.com/FirelyTeam/firely-net-sdk/discussions/1901](https://github.com/FirelyTeam/firely-net-sdk/discussions/1901)

This PR adds the first "base" properties to allow auto generating FiveWs attribute for resources.
It simply exposes a new Property called FiveWs which is a list of strings.

Once this is approved and merged, we will be able to add the codegen the ability to generate and utilize the FiveWs proportey.

### DIFF
--- a/src/Hl7.Fhir.Support.Poco/Introspection/PropertyMapping.cs
+++ b/src/Hl7.Fhir.Support.Poco/Introspection/PropertyMapping.cs
@@ -70,7 +70,7 @@ namespace Hl7.Fhir.Introspection
         public bool InSummary { get; private set; }
 
         /// <summary>
-        /// Weather the element has five ws
+        /// Five W's mappings of the element.
         /// </summary>
         public System.Collections.Generic.List<string> FiveWs { get; private set; }
 
--- a/src/Hl7.Fhir.Support.Poco/Introspection/PropertyMapping.cs
+++ b/src/Hl7.Fhir.Support.Poco/Introspection/PropertyMapping.cs
@@ -72,7 +72,7 @@ namespace Hl7.Fhir.Introspection
         /// <summary>
         /// Five W's mappings of the element.
         /// </summary>
-        public System.Collections.Generic.List<string> FiveWs { get; private set; }
+        public string[] FiveWs { get; private set; }
 
         /// <summary>
         /// Whether the element has a cardinality higher than 0.
--- a/src/Hl7.Fhir.Support.Poco/Introspection/PropertyMapping.cs
+++ b/src/Hl7.Fhir.Support.Poco/Introspection/PropertyMapping.cs
@@ -72,6 +72,9 @@ namespace Hl7.Fhir.Introspection
         /// <summary>
         /// Five W's mappings of the element.
         /// </summary>
+        /// <remarks>Each string in the array represents the exact element name of one the elements of the 
+        /// <c>FiveWs</c> pattern from http://hl7.org/fhir/fivews.html. Choice elements are spelled with the
+        /// [x] suffix, like <c>done[x]</c>. </remarks>
         public string[] FiveWs { get; private set; }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


ewoutkramer:
Thanks!  Is `string` the best representation here?  Or an enum with values for what, when,where,why,who?

itye-msft:
> Thanks! Is `string` the best representation here? Or an enum with values for what, when,where,why,who?
@ewoutkramer it looks like there many more values other than the Five W's.
Here is a short list we gathered quickly just to bring more clarity:

- class
- recorded
- witness
- why[x]
- identifier
- version
- status
- what[x]
- subject[x]
- subject
- done[x]
- actor
- where[x]
- planned
- grade
- author
- actor
- And more options...

It looks like this list is generated for each FHIR spec, and each spec may have different values there. Which mean this cannot be part of "common".
In addition, we followed the string pattern to keep it the same as other properties already present in the project, in order to work with the reflection system later.

ewoutkramer:
Ah, so this is a mapping to any of the elements in the pattern at http://hl7.org/fhir/fivews.html.

Yes - generating an enum per-version means that we could not use it from common, but (at least currently) none of the domain resources are in common.  It also depends a bit on the kind of use.  Is this for in a UI?  Or are you doing some kind of analysis on it?

itye-msft:
> Ah, so this is a mapping to any of the elements in the pattern at http://hl7.org/fhir/fivews.html.
> 
> Yes - generating an enum per-version means that we could not use it from common, but (at least currently) none of the domain resources are in common. It also depends a bit on the kind of use. Is this for in a UI? Or are you doing some kind of analysis on it?

@ewoutkramer the use case (for our case) would be to enforce a level of authorization  over entities. (e.g. a consent to view data of a patient). Having these mappings handy in the SDK would make the implementation a breeze.

Would you approve this PR?
Or would you like me to change/move the mapping to the specific R4 domain?
I thought that having it as string and use reflection, would allow it to be available to all domains.

ewoutkramer:
A string is then probably best. Shall we put `where[x]` there, or `where` ?

Will you update the code generator in cooperation with @ginocanessa?

itye-msft:
> A string is then probably best. Shall we put `where[x]` there, or `where` ?
> 
> Will you update the code generator in cooperation with @GinoCanessa?

In my team we thought where[x] is preferred in order to be aligned with the specs.
I already have a draft PR for the code gen in case this PR will be approved, with cooperation and consultation of Gino :)

ewoutkramer:
Nice! One more question: why is this an array?  Does this mean one element can play multiple role's in the "five W's"?

src/Hl7.Fhir.Support.Poco/Introspection/PropertyMapping.cs
```diff
@@ -69,6 +69,11 @@ public class PropertyMapping : IElementDefinitionSummary
         /// </summary>
         public bool InSummary { get; private set; }
 
+        /// <summary>
+        /// Five W's mappings of the element.
+        /// </summary>
```
ewoutkramer, line 74:
```suggestion
        /// </summary>
        /// <remarks>Each string in the array represents the exact element name of one the elements of the 
        /// <c>FiveWs</c> pattern from http://hl7.org/fhir/fivews.html. Choice elements are spelled with the
        /// [x] suffix, like <c>done[x]</c>. </remarks>
```

ewoutkramer: Maybe even add why it repeats.

ewoutkramer (changes requested):

itye-msft:
> Nice! One more question: why is this an array? Does this mean one element can play multiple role's in the "five W's"?

It's actually a good question. 
I looked at random generated code using this attribute (using the draft PR for codegen) and the only places I found more than one value in the array were the following similar to this one:
It's an array because it can certainly have multiple values, for example, this is a snippet of the "future" generated code :
```[FhirElement("subject", InSummary=true, Order=150, FiveWs= new string[] {"FiveWs.subject[x]","FiveWs.subject"})]
    [CLSCompliant(false)]
    [References("Patient","Group","Device","Location")]
    [DataMember]
    public Hl7.Fhir.Model.ResourceReference Subject
    {
      get { return _Subject; }
      set { _Subject = value; OnPropertyChanged("Subject"); }
    }
```
It looks like it's used for synonym phrases. But for sure there are places where more than one option is provided by the spec, and that's why an array is presented.

ewoutkramer (approved):

PR #169, opened 2021-11-24
Use existing ValidationAttributes in the new Json deserializer

ewoutkramer:
Updated the IsValidValue of most datatypes to make sure the attribute validation not only uses the regexes but uses semantic checks as well.


src/Hl7.Fhir.Support.Poco/Model/UnsignedInt.cs
```diff
@@ -0,0 +1,44 @@
...
+#nullable enable
+
+namespace Hl7.Fhir.Model
+{
+    public partial class UnsignedInt
+    {
+        /// <summary>
+        /// Checks whether the given literal is correctly formatted.
+        /// </summary>
+        public static bool IsValidValue(string value) => ElementModel.Types.Integer.TryParse(value, out var parsed) && parsed.Value > 0;
```
marcovisserFurore, line 40: `parsed.Value` can also be 0, see https://www.hl7.org/fhir/datatypes.html#unsignedInt

Actually, such rules should also be tested in unittests, I think.

src/Hl7.Fhir.Support.Poco/Serialization/PathStack.cs
```diff
@@ -18,66 +19,66 @@ namespace Hl7.Fhir.Serialization
...
         }
 
         /// <summary>
         /// Return the path. Note: in contained resources, this is just the path within the contained resource.
         /// </summary>
         public string GetPath()
         {
-            if (_pathStack.Count == 0)
-                throw new InvalidOperationException("No path present.");
+            if (_tops.Count == 0) return "$this";
```
marcovisserFurore, line 76: Should this not be an empty string? Stack is empty, then there is no path, right?

marcovisserFurore (changes requested):
One error in `unsignedInt`.

I was also wondering what the abbreviation `DAVE` stands for in 
```c#
using DAVE = Hl7.Fhir.Validation.CodedValidationException;
```

src/Hl7.Fhir.Support.Poco/Model/UnsignedInt.cs
ewoutkramer: Yes, we should.  In fact, I found another bug now I wrote them.

ewoutkramer:
> using DAVE = Hl7.Fhir.Validation.CodedValidationException;

This used to be

> using DAVE = Hl7.Fhir.Validation.DataAttributeValidationException;

But after a rename this direct correspondence got lost...

src/Hl7.Fhir.Support.Poco/Serialization/PathStack.cs
ewoutkramer: It's actually not clear what the path is.  Is it a path into a definition?  Is it a path into an instance?  I decided it would be the latter, and then it made sense to make it a FhirPath path (just like in OperationOutcome.issue).  Therefore, when you  are still in the root of an object, `$this` makes sense, more than an empty string (which debugs and display rather nastily).  

The *only* time this happens is when you validate an Element (not a resource), and immediately catch an error before going into the Element's properties.  The error will now display "At $this." instead of "At.".

marcovisserFurore (approved):

PR #172, opened 2021-11-30
Add FhirPath math for date(Time)/duration additions.

ewoutkramer:
Also applied changes pertaining to https://jira.hl7.org/browse/FHIR-28144 and the discussion in https://chat.fhir.org/#narrow/stream/179266-fhirpath/topic/Date.2FTime.20comparison.20vs.20equality

### DIFF
--- a/src/Hl7.FhirPath.Tests/Tests/BasicFunctionTests.cs
+++ b/src/Hl7.FhirPath.Tests/Tests/BasicFunctionTests.cs
@@ -189,7 +189,7 @@ namespace Hl7.FhirPath.Tests
             Assert.AreEqual(P.Date.Today(), scalar("today()"));
 
             // Check greater than
-            isB("today() < @" + P.Date.FromDateTimeOffset(DateTimeOffset.UtcNow.AddDays(1), includeOffset: false));
+            isB("today() < @" + P.Date.FromDateTimeOffset(DateTimeOffset.UtcNow.AddDays(2), includeOffset: false));
 
             // Check less than
             isB("today() > @" + P.Date.FromDateTimeOffset(DateTimeOffset.UtcNow.AddDays(-1), includeOffset: false));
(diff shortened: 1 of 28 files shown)


ewoutkramer:
This is basically PR #164, plus the changes to equality/comparisons due to  https://jira.hl7.org/browse/FHIR-28144 and the discussion in https://chat.fhir.org/#narrow/stream/179266-fhirpath/topic/Date.2FTime.20comparison.20vs.20equality

src/Hl7.FhirPath.Tests/Tests/BasicFunctionTests.cs
```diff
@@ -189,7 +189,7 @@ public void TestExpressionTodayFunction()
             Assert.AreEqual(P.Date.Today(), scalar("today()"));
 
             // Check greater than
-            isB("today() < @" + P.Date.FromDateTimeOffset(DateTimeOffset.UtcNow.AddDays(1), includeOffset: false));
+            isB("today() < @" + P.Date.FromDateTimeOffset(DateTimeOffset.UtcNow.AddDays(2), includeOffset: false));
```
marcovisserFurore, line 192: Why is this changed to +2 days instead of 1?

marcovisserFurore (changes requested):
See my remark

src/Hl7.FhirPath.Tests/Tests/BasicFunctionTests.cs
ewoutkramer: This test will otherwise intermittently fail depending on the time of day:  if you are east of GMT and it is already the next day where you are (compared to GMT), this means `today()` is already UTC + 1 day and is not `<`.  Doing `today() <= @.....` would also fix it.

ewoutkramer: (Brian found this bug, who is at GMT+10, so he runs into this much more often than we do)

marcovisserFurore (approved):

PR #174, opened 2021-12-20
Fix duplicate evaluation of scalar expressions

JoostSchipperheijn:
Hi Firely,

I noticed that scalar expressions are evaluated twice in the current implementation.
This has a performance penalty and can cause problems for custom functions that maintain state.

I fixed the problem and added a unit test to validate it.
The take(2) call composes the minimal number of objects 
required to determine whether the evaluation has 0, 1 or more results.

Can you have a look?

Thanks -- Joost Schipperheijn

### DIFF
--- a/src/Hl7.FhirPath.Tests/Functions/FunctionsTests.cs
+++ b/src/Hl7.FhirPath.Tests/Functions/FunctionsTests.cs
@@ -527,14 +527,13 @@ namespace HL7.FhirPath.Tests
             }
         }
 
-        private static int iterations = 0;
-
         /// <summary>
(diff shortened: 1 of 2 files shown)


marcovisserFurore:
Thanks @JoostSchipperheijn. We will investigate this. I've already run the [FhirPath ](https://github.com/FirelyTeam/firely-net-sdk/tree/develop-r4/src/Hl7.FhirPath.R4.Tests)tests for R4, and they succeeded as well with your solution.

ewoutkramer (approved):
Thanks @JoostSchipperheijn - nice unit test!

PR #185, opened 2022-02-14
Fix timeout bug in HttpClientRequester

cdaven:
`FhirClientSettings.Timeout` is a value in milliseconds, but was treated as seconds.

Note that this bug was introduced 2020-04-09, so it's possible that this has become the expected behaviour. See commit c87043532bac755f6884ed6b4d1d58ac00cc520e

Note that `WebClientRequester.ExecuteAsync()` handles `Timeout` correctly.


marcovisserFurore:
Hi @cdaven,

Thanks for reporting this issue and also give a solution for it by a pull request. 

I will pull it in the branch `release/4.0.0-beta2`, so it will come in version 4.0.0.

marcovisserFurore (approved):

PR #186, opened 2022-02-15
Validator checks if invariants don't result in empty anymore [common]

mmsmits:
fixes: https://github.com/FirelyTeam/firely-net-sdk/issues/1929


src/Hl7.FhirPath/FhirPath/CompiledExpression.cs
```diff
@@ -31,6 +30,18 @@ public static bool Predicate(this CompiledExpression evaluator, ITypedElement in
                 return result.Value;
         }
 
+        //Empty is considered false
+        public static bool IsTrue(this CompiledExpression evaluator, ITypedElement input, EvaluationContext ctx)
+        {
+            var result = evaluator(input, ctx).BooleanEval();
+
+            if (result == null)
```
ewoutkramer, line 38: I think we should get used to using `result is null` as the preferred syntax now.

src/Hl7.FhirPath/FhirPath/CompiledExpression.cs
```diff
@@ -1,8 +1,7 @@
-using System;
-using System.Collections.Generic;
-using System.Linq;
 using Hl7.Fhir.ElementModel;
```
ewoutkramer, line 1: Can you enable nullability & document the public members in this file please?

src/Hl7.FhirPath/FhirPath/TypedElementFPExtensions.cs
```diff
@@ -6,13 +6,9 @@
  * available at https://raw.githubusercontent.com/FirelyTeam/firely-net-sdk/master/LICENSE
```
ewoutkramer, line 6: Can you enable nullability & add documentation to public members please?

ewoutkramer (changes requested):

ewoutkramer (approved):

PR #194, opened 2022-04-12
Added Source Link

marcovisserFurore:
Added [Source Link](https://github.com/dotnet/sourcelink#source-link) for enabling anyone building NuGet libraries to provide source debugging.


ewoutkramer:
Is this for linking to the source from Visual Studio?  Interesting....you only seem to add a reference to a nuget, and we're not invoking anything....can adding a reference to nuget actually do/run something??

marcovisserFurore:
From the readme of the repo https://github.com/dotnet/sourcelink: 
> Source Link package is a development dependency, which means it is only used during build.

marcovisserFurore:
And this change gives us also nice green ticks at NuGet (notice also the package signing): 

image

The package signing was done in a previous release

mmsmits (approved):

PR #196, opened 2022-04-20
Use IssueComponent.Expression instead of IssueComponent.Location 

marcovisserFurore:
Resolves issue FirelyTeam/firely-net-sdk#1692

### DIFF
--- a/src/Hl7.Fhir.Support.Poco/Support/Issue.cs
+++ b/src/Hl7.Fhir.Support.Poco/Support/Issue.cs
@@ -42,7 +42,12 @@ namespace Hl7.Fhir.Support
             // Put numeric code + readable message into a CodeableConcept
             ic.Details = ToCodeableConcept(message);
 
-            if (path != null) ic.Location = new List<string> { path };
+            if (path is not null)
+            {
+                ic.Expression = new List<string> { path };
+                // IssueComponent.Location is deprecated but we still set this because of backwards compatibility.
+                ic.Location = new List<string> { path };
+            }
 
             return ic;
         }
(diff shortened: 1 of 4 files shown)


src/Hl7.Fhir.Support.Poco/Support/Issue.cs
```diff
@@ -42,7 +42,12 @@ public OperationOutcome.IssueComponent ToIssueComponent(string message, string p
             // Put numeric code + readable message into a CodeableConcept
             ic.Details = ToCodeableConcept(message);
 
-            if (path != null) ic.Location = new List<string> { path };
+            if (path is not null)
+            {
+                ic.Expression = new List<string> { path };
+                // IssueComponent.Location is deprecated but we still set this because of backwards compatibility.
```
ewoutkramer, line 48: Or should we change OperationOutcome.cs in common so that if you set Location, you set Expression and vice versa?

ewoutkramer (changes requested):
This is probably all good, but maybe we should change the property setters of Location/Expression to make sure both are always the same?

src/Hl7.Fhir.Support.Poco/Support/Issue.cs
brianpos: Just a tip that I've continued to use the location property for other error location styles (file positions line col, or questionnaire linkID, and having the fhirpath expression with the real one...)

brianpos: But I'm also happy for this to be copying the data in both properties here (as the location field was set by this code previously, so all should be fine)

ewoutkramer: > Just a tip that I've continued to use the location property for other error location styles
I was kinda expecting that yes, connecting the two together at the property level is also not backwards compatible...

So I found code setting Location in the STU3 PR (https://github.com/FirelyTeam/firely-net-sdk/pull/2048) where we then should set both Location and Expression (instead of making the setters for both the same).

ewoutkramer (approved):

PR #199, opened 2022-05-02
Corrected bugs in new faster serializers/deserializers

ewoutkramer:
* Avoided the "read too much" error in S.T.J,
* Added correct support (with a ConverterFactory) for creating converters for subclasses of Base.


ewoutkramer:
Note: this bugfix requires a change to the public interface of 4.0.0-beta2, since FhirJsonConverter has been changed to FhirJsonConverter<T> (and in fact, we need to use FhirJsonConverterFactory instead).

PR #218, opened 2022-07-29
1688 - Add functionality to have "#" resolve to container

ewoutkramer:
Patches the Resolve() on ScopedNode to be able to resolve to the container Parent.

### DIFF
--- a/src/Hl7.Fhir.ElementModel/ScopedNode.cs
+++ b/src/Hl7.Fhir.ElementModel/ScopedNode.cs
@@ -25,24 +25,28 @@ namespace Hl7.Fhir.ElementModel
             public string? Id;
             public IEnumerable<ScopedNode>? ContainedResources;
             public IEnumerable<BundledResource>? BundledResources;
+
+            public string? InstanceUri;
         }
 
         private readonly Cache _cache = new();
 
         public readonly ITypedElement Current;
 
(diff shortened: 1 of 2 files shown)


src/Hl7.Fhir.ElementModel/ScopedNode.cs
```diff
@@ -25,24 +25,28 @@ private class Cache
...
+
+            public string? InstanceUri;
         }
 
         private readonly Cache _cache = new();
 
         public readonly ITypedElement Current;
 
-        public ScopedNode(ITypedElement wrapped)
+        public ScopedNode(ITypedElement wrapped, string? instanceUri = null)
```
marcovisserFurore, line 36: Is this not breaking the interface? I see there is a nullable second parameter, but the definition of this constructor has changed. Should it not be better to add a second constructor:
```c#
public ScopedNode(ITypedElement wrapped, string instanceUri )
{
   // etc...
}

ewoutkramer: That's equivalent - the second parameter is not only nullable, but also optional (`= null`).

ewoutkramer: BTW: It is not *binary* compatible, but that was never a goal...

marcovisserFurore (approved):

PR #222, opened 2022-08-03
2018 Make sure repeats() does not hang when given a constant argument

ewoutkramer:
Updated the repeat() algorithm to enforce uniqueness of newly added items.

### DIFF
--- a/src/Hl7.FhirPath/FhirPath/Expressions/SymbolTableInit.cs
+++ b/src/Hl7.FhirPath/FhirPath/Expressions/SymbolTableInit.cs
@@ -346,7 +346,10 @@ namespace Hl7.FhirPath.Expressions
                     newContext.SetIndex(ElementNode.CreateList(index));
                     index++;
 
-                    newNodes.AddRange(lambda(newContext, InvokeeFactory.EmptyArgs));
+                    var candidates = lambda(newContext, InvokeeFactory.EmptyArgs);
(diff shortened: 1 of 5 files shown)


src/Hl7.FhirPath/FhirPath/Expressions/SymbolTableInit.cs
```diff
@@ -346,7 +346,10 @@ private static IEnumerable<ITypedElement> runRepeat(Closure ctx, IEnumerable<Inv
                     newContext.SetIndex(ElementNode.CreateList(index));
                     index++;
 
-                    newNodes.AddRange(lambda(newContext, InvokeeFactory.EmptyArgs));
+                    var candidates = lambda(newContext, InvokeeFactory.EmptyArgs);
```
marcovisserFurore, line 349: Could you not use the Linq method Except? So:
```c#
 newNodes.AddRange(candidates.Except(fullResult, EqualityOperators.TypedElementEqualityComparer));
```

marcovisserFurore (changes requested):
See my suggestion

src/Hl7.FhirPath/FhirPath/Expressions/SymbolTableInit.cs
ewoutkramer: Yes!

marcovisserFurore (approved):

PR #228, opened 2022-08-25
2192 - Make Response Headers case insensitive

### LINKED ISSUE #2192: Case-sensitivity for custom HTTP headers in response
### Discussed in https://github.com/FirelyTeam/firely-net-sdk/discussions/2107

<div type='discussions-op-text'>

<sup>Originally posted by **cdaven** June  1, 2022</sup>
The HTTP specification says that headers are **case insensitive** (see [RFC 2616, section 4.2](https://www.rfc-editor.org/rfc/rfc2616#section-4.2)). .NET parses the basic headers and puts them in the `HttpResponseMessage.Headers` object, so we don't have to think about this. However, all custom headers are copied to the `EntryResponse` object by `HttpToEntryExtensions.SetHeaders()` into a **case-sensitive** `Dictionary`.

You can see what I mean in the unit test `TestToTypedEntryResponse()`, and change this line:

```csharp
Headers = new Dictionary<string, string>() { { "Test-key", "Test-value" } },
```

Change e.g. `"Test-key"` to `"test-key"`, and the test will fail. Is this intended or a bug?

I thought we could make this easy by making `EntryResponse.Headers` case-insensitive in the constructor:

```csharp
public EntryResponse()
{
    Headers = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
}
```

But since `Headers` is replaced in e.g. `ToTypedEntryResponse()`, that would have to be refactored as well.

What do you think? The Swedish NLL implementation has suddenly announced that they will start using lowercase headers only.</div>

marcovisserFurore:
The headers in `EntryResponse` are now treated as case insensitive headers.

Resolves FirelyTeam/firely-net-sdk#2192

### DIFF
--- a/src/Hl7.Fhir.Support.Poco.Tests/HttpToEntryExtensionsTests.cs
+++ b/src/Hl7.Fhir.Support.Poco.Tests/HttpToEntryExtensionsTests.cs
@@ -1,0 +1,43 @@
+﻿using FluentAssertions;
+using Hl7.Fhir.Rest;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using System;
+using System.Collections.Generic;
+using System.Net;
+using System.Net.Http;
+using System.Reflection;
+
+namespace Hl7.Fhir.Support.Poco.Tests
+{
+
+    [TestClass]
+    public class HttpToEntryExtensionsTests
+    {
+
+        [TestMethod]
+        public void ResponseHeadersCaseInsensitiveTest()
+        {
+            var headers = new WebHeaderCollection
+            {
+                { "LOWERCASE", "value" }
+            };
+
+            var mock = new Mock<HttpWebResponse>();
+            mock.Setup(c => c.Headers).Returns(headers);
+            var httpWebResponse = mock.Object;
+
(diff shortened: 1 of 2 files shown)


brianpos:
I'd love to see this test also check what happens when there are multiple values in the header.

The HttpRequest and HttpResponse use this class for headers, which isn't a string to string dictionary, it can have multiple values.
public partial interface IHeaderDictionary : IDictionary<string, StringValues>

This is how I've read the headers in my project
`Request.Headers.Select(h => new KeyValuePair<string, IEnumerable<string>>(h.Key, h.Value)),`
https://github.com/brianpos/fhir-net-web-api/blob/874980ee307d47062f5d17bcd99760219a86a69c/src/Hl7.Fhir.WebApi.AspNetCore/FhirController.cs#L116
And I'm using an enumeration of string to string enumeration ley value pairs

marcovisserFurore:
Well spotted @brianpos! My test was only for the old `HttpWebResponse`. Nowadays we have the `HttpResponseMessage`. I will create an extra test for that as well.

ewoutkramer (approved):

PR #231, opened 2022-08-31
Fix NullRefEx on FhirDateTime.TryToDateTimeOffset 

GitGuruGangsta:
When FhirDateTime is null and you try to invoke TryToDatetimeOffset it throws a NullReferenceException.

Try-Methods shouldn't do that and instead respond with false.

[#2218](https://github.com/FirelyTeam/firely-net-sdk/issues/2218)

### DIFF
--- a/src/Hl7.Fhir.Support.Poco/Model/FhirDateTime.cs
+++ b/src/Hl7.Fhir.Support.Poco/Model/FhirDateTime.cs
@@ -143,7 +143,7 @@ namespace Hl7.Fhir.Model
         {
             dto = default;
 
-            if (this.Value == null)
+            if (this == null || this.Value == null)
                 return false;
 
             if (tryGetTimeSpan(out var timespan))


ewoutkramer:
Is this the case you are concerned with?

``` csharp
FhirDateTime nullFDT = null;
var success = nullDDT.TryToDateTimeOffset(out var dto);
dto.Dump();
```

In that case, this fix will not fix it. `TryToDateTimeOffset()` is not an extension method, nor is it a static method - `this` will never be null - the null reference exception will happen at the callsite.

If we need to fix this, `TryToDateTimeOffset()` will need to be changed into an extension method, so it takes "this" as the first parameter:

``` csharp
public static bool TryToDateTimeOffset(this FhirDateTime fdt, out DateTimeOffset dto)
{
    // now we check fdt against null
    if(fdt is null) return false;
}
```

But this is pretty unusual, e.g. this `Try` function from the .NET BCL will also throw:

``` csharp
Dictionary<string,string> x = null;
x.TryGetValue("bla", out var _);
```

So, I don't really see the point.

PR #232, opened 2022-09-06
Fix index out of bounds error when trying to parse illegal history resource identifier [common]

### LINKED ISSUE #2190: ResourceIdentity error with reference XXX/_history/1
When creating a new ResourceIdentity based on `XXX/_history/1`, the SDK throws a `System.IndexOutOfRangeException`, because the `_history` component isn't in the correct place (or either the resource type or version are missing). 

Expected behaviour: Don't crash, but return a ResourceIdentity without any components (because it's an invalid identity).

mmsmits:
fixes: FirelyTeam/firely-net-sdk/issues/2190

### DIFF
--- a/src/Hl7.Fhir.Support.Tests/Rest/TestResourceIdentifier.cs
+++ b/src/Hl7.Fhir.Support.Tests/Rest/TestResourceIdentifier.cs
@@ -1,0 +1,265 @@
+﻿/* 
+ * Copyright (c) 2014, Firely (info@fire.ly) and contributors
+ * See the file CONTRIBUTORS for details.
+ * 
+ * This file is licensed under the BSD 3-Clause license
+ * available at https://raw.githubusercontent.com/FirelyTeam/firely-net-sdk/master/LICENSE
+ */
+
+using FluentAssertions;
+using Hl7.Fhir.Rest;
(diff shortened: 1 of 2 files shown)


marcovisserFurore (commented):
We discussed this morning that the `TestResourceIdentifier` can also be moved to Common

src/Hl7.Fhir.Support.Tests/Rest/TestResourceIdentifier.cs
```diff
@@ -0,0 +1,265 @@
...
+            Assert.IsFalse(ResourceIdentity.IsRestResourceIdentity("urn:oid:1.2.3.4.5"));
+            Assert.IsFalse(ResourceIdentity.IsRestResourceIdentity("XXX/_history/1"));
+        }
+
+
+        [DataTestMethod]
+        [DataRow("XXX/1/_history/1", "1")]
+        [DataRow("XXX/_history/1", null)]
+        [DataRow("_history/1", null)]
+        public void TestParsingVersionId(string uri, string expectedVersionId)
```
marcovisserFurore, line 259: Very nice!

marcovisserFurore (approved):

PR #233, opened 2022-09-13
Permit passing the ExceptionHandler through the settings object

brianpos:
This ExceptionHandler will now permit configuration of a handler to intercept any exceptions during parsing to be able to enhance processing/error reporting.


brianpos:
This is the change for https://github.com/FirelyTeam/firely-net-sdk/issues/2222

marcovisserFurore (changes requested):
Thanks for the PR Brian, but the buildserver found some unit test failures. So I cannot pull it. Also I miss some unit test for this change, so the new settings is not tested.

marcovisserFurore (approved):

PR #239, opened 2022-10-28
#2268 include accessors and additional constructor for Canonical components

brianpos:
This implements the functionality requested by [#2268 (in the FirelyTeam SDK repo)](https://github.com/FirelyTeam/firely-net-sdk/issues/2268)
It includes a unit test that verifies all the behaviours that I would have expected.

### DIFF
--- a/src/Hl7.Fhir.Support.Poco/Model/Canonical.cs
+++ b/src/Hl7.Fhir.Support.Poco/Model/Canonical.cs
@@ -1,33 +1,35 @@
 ﻿/*
   Copyright (c) 2011+, HL7, Inc.
   All rights reserved.
-  
-  Redistribution and use in source and binary forms, with or without modification, 
+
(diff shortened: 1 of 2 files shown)


brianpos:
@ewoutkramer - from your comments in slack... 
> Yes, we'd need to keep an "original" value of the objectvalue - and compare it against the actual objectvalue to detect changes. That's yuck, but right now you have three (partial) copies of the same parsing logic in each propery, which is executed each time - I don't like that either.

However the code in here isn't the same, it's the minimal amount that you need to read the specific value being requested. And doesn't have any issues with being in sync with the base value (if were in shadow properties) - This code could even be done with extension methods!

ewoutkramer:
Moved to https://github.com/FirelyTeam/firely-net-sdk/pull/2358, since this common repo is not in use anymore starting from SDK 5.0.

PR #241, opened 2022-11-04
Feature/big common

marcovisserFurore:
- Added Bindable attribute
- `ModelInspector` implements now `IModelInfo`


src/Hl7.Fhir.Support.Poco/Model/ModelInfoExtensions.cs
```diff
@@ -0,0 +1,21 @@
+﻿using Hl7.Fhir.Introspection;
+using Hl7.Fhir.Model;
+using System;
+
+#nullable enable
+
+namespace Hl7.Fhir.Support.Poco.Model
+{
+    public class ModelInfoExtensions
```
ewoutkramer: These are not extensions....I think the intent was to make these extension methods on IModelInfo, so you also don't need to do the `ModelInspector.ForAssembly()`, you would just access the ModelInspector property from IModelInfo (if it does exist?) 

Then again, are these not simply members of IModelInfo?  I realize I don't understand why we need these extension methods.......  is it to get to model info stuff without having a ModelInspector?  Then, should we not change the name of this class since they are not extension methods?

ewoutkramer (changes requested):

ewoutkramer (approved):

PR #248, opened 2022-11-17
Improved null handling in PackageSource

ewoutkramer:
Added nullability to IConformanceResource and fixed related errors.

### DIFF
--- a/src/Hl7.Fhir.Support.Poco/Specification/Source/CommonFhirPackageSource.cs
+++ b/src/Hl7.Fhir.Support.Poco/Specification/Source/CommonFhirPackageSource.cs
@@ -52,7 +52,7 @@ namespace Hl7.Fhir.Specification.Source
 
         }
 
-        private async static Task<PackageManifest> initialize(string path, string name, string version, string author, string description, string[] dependencies)
+        private static async Task<PackageManifest> initialize(string path, string name, string version, string author, string description, string[] dependencies)
         {
             if (!Directory.Exists(path)) Directory.CreateDirectory(path);
             var project = new FolderProject(path);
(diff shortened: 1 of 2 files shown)


src/Hl7.Fhir.Support.Poco/Specification/Source/CommonFhirPackageSource.cs
```diff
@@ -276,21 +276,20 @@ public IEnumerable<string> ListResourceUris(string? filter = null)
         }
 
         ///<inheritdoc/>
-        public async Task<IEnumerable<Resource>?> FindConceptMaps(string? sourceUri = null, string? targetUri = null)
+        public async Task<IEnumerable<Resource>> FindConceptMaps(string? sourceUri = null, string? targetUri = null)
```
marcovisserFurore, line 279: This is a breaking change, right?

marcovisserFurore: When yes, then we should add this change in https://github.com/FirelyTeam/firely-net-sdk/wiki/Breaking-changes-in-5.0

marcovisserFurore (approved):
See my breaking change question.

src/Hl7.Fhir.Support.Poco/Specification/Source/CommonFhirPackageSource.cs
ewoutkramer: I think so - but it's like the trick we use at HL7 now and then: the intention of these interface methods was to return an empty collection (the original implementers of this interface do so) - so this was a bug ;-)

When I thought about the consequences I figured it might mean there are now some null checks that will never be hit, but all the logic working on lists will probably still function perfectly fine.

I'll mark this PR as breaking, and update the wiki!

PR #249, opened 2022-11-21
FhirClient to common

marcovisserFurore:
Move the following resources from big common to common:
- `Binary`
- `Bundle`
- `Signature`

Move FhirClient from the satellites to common.

Created an instantiable `CommonFhirClient`, which can be used in multi-version solutions.

Todo:
- Change codegen to reflect these changes.


src/Hl7.Fhir.Support.Poco.Tests/NewPocoSerializers/SummaryFilterTests.cs
```diff
@@ -16,7 +16,7 @@ public void BundleFilter()
             var epf = new ElementPrefixFilter("elem");
             var bf = new BundleFilter(epf);
 
-            var inspector = ModelInspector.ForAssembly(typeof(TestBundle).Assembly);
+            var inspector = ModelInspector.ForAssembly(typeof(TestPatient).Assembly);
 
             bf.EnterObject(null, inspector.FindClassMapping("Bundle"));
```
ewoutkramer, line 21: The reason this was typeof(TestBundle), is that we navigate here into a Bundle. Maybe this should be `ForAssembly(typeof(Bundle))` now?

src/Hl7.Fhir.Support.Poco/Serialization/CommonFhirJsonSerializer.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        public CommonFhirJsonSerializer(ModelInspector modelInspector, SerializerSettings settings = null) : base(modelInspector, settings)
+        {
+        }
+
+        private FhirJsonSerializationSettings buildFhirJsonWriterSettings() =>
+            new() { Pretty = Settings.Pretty, AppendNewLine = Settings.AppendNewLine };
+
+        /// <inheritdoc cref="SerializeToStringAsync(Base, SummaryType, string[])" />
+        public string SerializeToString(Base instance, SummaryType summary = SummaryType.False, string[] elements = null) =>
```
ewoutkramer, line 28: Can we push these methods down to the base?

ewoutkramer (changes requested):

ewoutkramer (approved):

PR #251, opened 2022-11-22
Unify terminology interfaces

ewoutkramer:
This PR splits up the big ITerminologyService interface into several smaller and more cohesive ones. 

I'd like the reviewers opinion on whether this makes sense and whether the split up is meaningful.

Note that there is a breaking change in this PR since (to make the interface useable across R3 and R4+) I needed to push `ValidateCodeParameters` to common, which meant changing the type of a property from ValueSet to Resource.


src/Hl7.Fhir.Support.Poco/Specification/Terminology/ValidateCodeParameters.cs
```diff
@@ -0,0 +1,179 @@
...
+        /// </summary>
+        public FhirUri Url { get; private set; }
+        /// <summary>
+        /// The context of the value set, so that the server can resolve this to a value set to validate against.
+        /// </summary>
+        public FhirUri Context { get; private set; }
+        /// <summary>
+        /// The value set is provided directly as part of the request.
+        /// </summary>
+        public Resource ValueSet { get; private set; }
```
ewoutkramer, line 108: Note that this used to be `ValueSet` - so this is a breaking change.

src/Hl7.Fhir.Support.Poco/Specification/Terminology/ITerminologyService.cs
```diff
@@ -22,29 +22,37 @@ public interface ITerminologyService
...
         /// <exception cref="FhirOperationException">Thrown when the terminology service encounters an error</exception>
-        Task<Parameters> CodeSystemValidateCode(Parameters parameters, string id = null, bool useGet = false);
+        /// <remarks>This function corresponds to the <seealso href="http://hl7.org/fhir/codesystem-operation-subsumes.html">$subsumes</seealso> operation.</remarks>
+        Task<Parameters> Subsumes(Parameters parameters, string id = null, bool useGet = false);
+    }
 
+    /// <summary>
+    /// Subset of the FHIR terminology service (http://hl7.org/fhir/terminology-service.html) that concerns lookup and validation of single codes.
+    /// </summary>
+    public interface ICodeSystemTerminologyService : ICodeValidationTerminologyService
```
marcovisserFurore:
```suggestion
    public interface ICodeSystemTerminologyService
```

marcovisserFurore: `ICodeSystemTerminologyService` should not be dependent on `ICodeValidationTerminologyService`

marcovisserFurore (changes requested):

marcovisserFurore (approved):

PR #254, opened 2022-12-05
Remove unnecessary dependency on SymbolTable from ExpressionVisitor

ewoutkramer:
The `ExpressionVisitor` used a `SymbolTable` in its `Accept` method, which was too specific and unnecessary.


src/Hl7.FhirPath/FhirPath/Expressions/EvaluatorVisitor.cs
```diff
@@ -16,33 +16,41 @@ namespace Hl7.FhirPath.Expressions
 {
     internal class EvaluatorVisitor : FP.ExpressionVisitor<Invokee>
     {
-        public override Invokee VisitConstant(FP.ConstantExpression expression, SymbolTable scope)
+        public SymbolTable Symbols { get; }
```
marcovisserFurore, line 19: Is it necessary that this property `Symbols` is public?

marcovisserFurore: Discussed together: this property is part of a internal class, so it is not very important.

marcovisserFurore (approved):

